Issue #2, opened 2020-10-04
Error when selecting issue from list


sunjon:
My Telescope issue picker config based on the example in the README now generates an error when selecting an issue:

```vim
Error executing vim.schedule lua callback: ...share/nvim/site/pack/packer/start/octo.nvim/lua/octo
.lua:427: Vim(lua):E5108: Error executing lua ...share/nvim/site/pack/packer/start/octo.nvim/lua/o
cto.lua:136: Key not found: labels                                                                
Press ENTER or type command to continue   
```

pwntester:
No description provided.

pwntester:
I improved your caching integration since it was caching to the same location for all repos:

```
local function issues_sujon(repo)

  if not vim.g.octo_last_results then vim.g.octo_last_results = {} end
  if not vim.g.octo_last_updatetime then vim.g.octo_last_updatetime = {} end

  local cache_timeout = 300 -- 5 min cache
  local current_time = os.time()
  local next_check
  if vim.g.octo_last_updatetime[repo] ~= nil then
    next_check = tonumber(vim.g.octo_last_updatetime[repo]) + cache_timeout
  else
    next_check = 0 -- now
  end

  local results = {}
  if current_time > next_check then
    local resp = require'octo'.get_repo_issues(repo, {})
    for _,i in ipairs(resp.issues) do
      table.insert(results, {
        number = i.number;
        title = i.title;
      })
    end
    local last_results = vim.api.nvim_get_var('octo_last_results')
    last_results[repo] = results
    vim.api.nvim_set_var('octo_last_results', last_results)

    local last_updatetime = vim.api.nvim_get_var('octo_last_updatetime')
    last_updatetime[repo] = current_time
    vim.api.nvim_set_var('octo_last_updatetime', last_updatetime)
  else
    results = vim.g.octo_last_results[repo]
  end

  local make_issue_entry = function(result)
    return {
      valid = true;
      entry_type = make_entry.types.GENERIC;
      value = tostring(result.number);
      ordinal = tostring(result.number);
      display = string.format('#%d - %s', result.number, result.title);
    }
  end

  local custom_mappings = function(prompt_bufnr, map)
    local run_command = function()
      local selection = actions.get_selected_entry(prompt_bufnr)
      actions.close(prompt_bufnr)
      local cmd = string.format([[ lua require'octo'.get_issue('%s', '%s') ]], selection.value, repo)
      vim.cmd [[stopinsert]]
[...]

PR #3, opened 2020-10-05
Fix issue #2 and allow for opts customization

sunjon:
This attempts to fix [issue #2 ](https://github.com/pwntester/octo.nvim/issues/2).

The default finders accept `opts` as a parameter. This matches that behaviour and ensures that the opts var is passed to functions that require it).

Secondly, autocommands are disabled when creating the issues buffer, as that triggers functions that try to read `b:` vars that haven't been set yet.

### DIFF
--- a/lua/octo.lua
+++ b/lua/octo.lua
@@ -428,7 +428,7 @@ local function create_issue_buffer(issue, repo)
 	local extmarks = {}
 
 	-- create buffer
-	api.nvim_command(format('e octo://%s/%s', repo, number))
+	api.nvim_command(format('noautocmd e octo://%s/%s', repo, number))
 	local bufnr = api.nvim_get_current_buf()
 
 	-- delete extmarks
--- a/lua/octo.lua
+++ b/lua/octo.lua
@@ -624,7 +624,7 @@ local function get_url(url, params)
 	return url
 end
 
-local function get_repo_issues(repo, query_params)
+local function get_repo_issues(opts, repo, query_params)
 
 	-- this function should be used by fuzzy pickers as a source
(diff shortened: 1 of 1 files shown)


lua/octo.lua
```diff
@@ -428,7 +428,7 @@ local function create_issue_buffer(issue, repo)
 	local extmarks = {}
 
 	-- create buffer
-	api.nvim_command(format('e octo://%s/%s', repo, number))
+	api.nvim_command(format('noautocmd e octo://%s/%s', repo, number))
```
pwntester, line 431: nice catch, this will prevent calling `show_details_win` twice

lua/octo.lua
```diff
@@ -624,7 +624,7 @@ local function get_url(url, params)
 	return url
 end
 
-local function get_repo_issues(repo, query_params)
+local function get_repo_issues(opts, repo, query_params)
```
pwntester, line 627: why? this function is not expecting any opts

lua/octo.lua
```diff
@@ -711,7 +711,7 @@ end
 --     curl.request(issues_url, opts, choose_issue)
 -- end
 
-local function get_issue(number, repo)
+local function get_issue(opts, number, repo)
```
pwntester, line 714: why? this function is not expecting any opts

sunjon:
I believe the opts were residue of me debugging the autocommand strangeness.
ie. functions called by the premature autocommands were throwing errors related to opts, and I was making them available until I got to the root cause.

Saw you committed the `noautocmd` fix separately, so closing this.

PR #6, opened 2020-12-04
New GH CLI wrapper

pwntester:


weilbith:
Alright, so I gave the plugin a first try since you now started to implement this feature. Pretty hyped. :grimacing: 

Unfortunately I came not far. I used the `ListIssues` command. This works great in combination with telescope. But when I want to open an issue, I get an empty buffer and this error message: `stdout { '{"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/issues#get-an-issue"}' }`.
I can confirm that I'm logged in. The output of `gh auth status` is:

```sh
github.com
  ✓ Logged in to github.com as weilbith (~/.config/gh/hosts.yml)
  ✓ Git operations for github.com configured to use ssh protocol.
  ✓ Token: *******************
```

pwntester:
uhm weird, what is the name of the buffer opened? what is the repo/issue you are trying to open?

weilbith:
The buffer is named like the number of the issue. The repo is just one from work (just because it has thousands of issues :shrug: ). I could also just test it with this one. But it is open source. So theoretically the authentication is theoretically not even necessary. At least not for showing.

pwntester:
> The buffer is named like the number of the issue.

Is that the output from `bufname` or how its listed by `ls!`?

weilbith:
Sorry that was just the named displayed in my tabline. So the relevant output from `:ls!` should be that:

```vim
 11u     "term://~/projects/brainbot-projects/repositories/raiden/light-client/raiden-dapp//792503:GIT_PAGER=less gh issue view 2435" line 1
 12 %a   "github://weilbith/light-client/2435" line 1
```

pwntester:
The buffer name seems correct, what is the output of `gh api repos/weilbith/light-client/issues/2435` in the command line?

weilbith:
Oh wait. I think I know what the issue might be related to. The get remote called `origin` is my fork. The "original" repo is called `upstream`. On `origin` there are no issues. But why does the listing of the issues work correctly?

pwntester:
Thats weird, the only difference I can think of is that in the `api` call you are explicitly setting the repo while in `gh issue list` the repo is inferred. Can you also try the following to force `api` to do a similar infer:

```
gh api repos/:owner/:repo/issues/2435
```

pwntester:
Please try `:ListIssues repo=weilbith/light-client` and then open the issue to avoid using the upstream repo implicitly

pwntester:
Merging since seems to work ok. I made some changes to make commands more consistent. Use `:ListIssues weilbith/light-client` instead

PR #43, opened 2020-12-26
Add lua-date to format dates

### LINKED ISSUE #38: Support for lua-date
I think you saw how I used `lua-date` in my plugin.
I would be happy to add this here.
It would format all dates to be more human-readable, or fallback silently to how it is now if `lua-date` is not installed.

lukas-reineke:
I tried to only copy the date stuff that's needed, but the code is not really modular.
Everything depends on everything..
It's not that big, I think it's acceptable to just include the whole thing. What do you think?

Also there should probably be a way to configure the formatting

resolve #38

### DIFF
--- a/lua/octo/util.lua
+++ b/lua/octo/util.lua
@@ -160,9 +160,9 @@ function M.update_reactions_at_cursor(bufnr, cursor, reactions, reaction_line)
   api.nvim_buf_set_var(bufnr, "reaction_line", reaction_line)
 end
 
-function M.format_date(to_format)
+function M.format_date(date_string)
   local time_bias = date():getbias() * -1
-  return date(to_format):addminutes(time_bias):fmt(vim.g.octo_date_format)
+  return date(date_string):addminutes(time_bias):fmt(vim.g.octo_date_format)
 end
 
 return M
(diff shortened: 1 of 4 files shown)


lua/octo/util.lua
```diff
@@ -143,4 +144,9 @@ function M.update_reactions_at_cursor(bufnr, cursor, reactions, reaction_line)
   api.nvim_buf_set_var(bufnr, "reaction_line", reaction_line)
 end
 
+function M.format_date(to_format)
```
pwntester, line 147: should we rename `to_format` to `date`?

lua/octo/util.lua
```diff
@@ -143,4 +144,9 @@ function M.update_reactions_at_cursor(bufnr, cursor, reactions, reaction_line)
...
+  local time_bias = date():getbias() * -1
+  return date(to_format):addminutes(time_bias):fmt("%Y %b %d %I:%M %p %Z")
```
pwntester, line 149: make format configurable with that format as the default value? we can use `vim.g.octo_date_format` for it

pwntester (changes requested):
Thanks, some minor changes requested

lua/octo/util.lua
lukas-reineke: not happy about `to_format` either, but `date` is already the lua-date import.
Maybe `date_string`? `iso8601`?

pwntester: `date_string` lgtm 👍🏼

pwntester:
thanks!

Issue #12, opened 2020-12-11
Add @people completion


pwntester:
`https://github.com/suggestions/issue/710584628?repository=telescope.nvim&user_id=nvim-lua&mention_suggester=1`
    - [repository contributors](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#list-repository-contributors)
    - issue author
    - comment authors
    - users already mentioned in comments or description

pwntester:
No description provided.

pwntester:
fixed in #56 by grabbing issue participants from graphql api

Antonio-Bennett:
This is a test using Octo.nvim lol

PR #48, opened 2020-12-28
Add @People mentioning to PR view

PatOConnor43:
### Description
This PR adds the ability to @mention people in the PR view. Related to #12. 

### Implementation
This works by accumulating a list of users from a couple different sources.
These include:
  - The author of the issue
  - The authors of existing comments
  - The contributors to the repo

This list of users is currently only used to @mention people so the api call to
fetch collaborators is started in the background without waiting for it to
complete. The accumulated list is then attached to the issue buffer under the
key `taggable_users`.

If the completion base starts with `@` inside the completion function, the
entries will be built from the `taggable_users` that should be attached to the
buffer.

I also added keymaps in insert mode for `@` and `#` to make them immediately
call the completion menu.

### Future Work
#12 mentions including users who were mentioned in existing comments or the
description. This is not included in this PR but I thought this could be a good
first step.

Thanks for taking the time to review.

### DIFF
--- a/lua/octo/completion.lua
+++ b/lua/octo/completion.lua
@@ -49,7 +49,7 @@ local function get_repo_issues(repo, params)
 end
 
 function M.issue_complete(findstart, base)
-  -- the complete-functions
+  -- :help complete-functions
   if findstart == 1 then
     -- findstart
     local line = api.nvim_get_current_line()
--- a/lua/octo/completion.lua
+++ b/lua/octo/completion.lua
@@ -11,12 +11,13 @@ local M = {}
 local function get_url(url, params)
   url = url .. "?foo=bar"
   for k, v in pairs(params) do
-    url = format("%s\\&%s=%s", url, k, v)
+    url = format("%s&%s=%s", url, k, v)
   end
   return url
 end
 
 local function get_repo_issues(repo, params)
+  params = params or {}
   local query_params = {
     state = params.state or "open",
     per_page = params.per_page or 50,
@@ -48,19 +49,25 @@ local function get_repo_issues(repo, params)
 end
 
 function M.issue_complete(findstart, base)
-  -- :he complete-functions
+  -- the complete-functions
   if findstart == 1 then
     -- findstart
     local line = api.nvim_get_current_line()
     local pos = vim.fn.col(".")
-    local i, j = 0
+    local i, j = 0, 0
     while true do
       i, j = string.find(line, "#(%d*)", i + 1)
       if i == nil then
-        break
+        i, j = 0, 0
+        i, j = string.find(line, "@(%w*)", i + 1)
+        if i == nil then
+          break
+        end
       end
       if pos > i and pos <= j + 1 then
-        return i
+        -- I think subtracting 1 is necessary to include the first character
+        -- since lua is 1 indexed
+        return i - 1
       end
     end
     return -2
@@ -68,18 +75,25 @@ function M.issue_complete(findstart, base)
     local repo = api.nvim_buf_get_var(0, "repo")
     local issues = get_repo_issues(repo)
     local entries = {}
-    for _, i in ipairs(issues) do
-      if vim.startswith(tostring(i.number), base) then
-        table.insert(
-          entries,
-          {
-            word = tostring(i.number),
-            abbr = format("#%d", i.number),
-            menu = i.title
-          }
-        )
+    if vim.startswith(base, "@") then
+      local users = api.nvim_buf_get_var(0, "taggable_users") or {}
+      for _, user in pairs(users) do table.insert(entries, {word = format("@%s", user), abbr = user}) end
+    else if vim.startswith(base, "#") then
+        for _, i in ipairs(issues) do
+          if vim.startswith(tostring(i.number), base) then
+            table.insert(
+              entries,
+              {
+                word = tostring(i.number),
+                abbr = format("#%d", i.number),
+                menu = i.title
+              }
+              )
+          end
+        end
       end
     end
     return entries
   end
 end
+return M
--- a/lua/octo/completion.lua
+++ b/lua/octo/completion.lua
@@ -72,20 +72,20 @@ function M.issue_complete(findstart, base)
     end
     return -2
   elseif findstart == 0 then
-    local repo = api.nvim_buf_get_var(0, "repo")
-    local issues = get_repo_issues(repo)
     local entries = {}
     if vim.startswith(base, "@") then
       local users = api.nvim_buf_get_var(0, "taggable_users") or {}
       for _, user in pairs(users) do table.insert(entries, {word = format("@%s", user), abbr = user}) end
     else if vim.startswith(base, "#") then
+        local repo = api.nvim_buf_get_var(0, "repo")
+        local issues = get_repo_issues(repo)
         for _, i in ipairs(issues) do
-          if vim.startswith(tostring(i.number), base) then
+          if vim.startswith('#'..tostring(i.number), base) then
             table.insert(
               entries,
               {
-                word = tostring(i.number),
-                abbr = format("#%d", i.number),
+                abbr = tostring(i.number),
+                word = format("#%d", i.number),
                 menu = i.title
               }
               )
--- a/lua/octo/completion.lua
+++ b/lua/octo/completion.lua
@@ -1,52 +1,7 @@
-local gh = require("octo.gh")
 local format = string.format
 local api = vim.api
-local json = {
-  parse = vim.fn.json_decode,
-  stringify = vim.fn.json_encode
-}
 
 local M = {}
-
-local function get_url(url, params)
-  url = url .. "?foo=bar"
-  for k, v in pairs(params) do
-    url = format("%s&%s=%s", url, k, v)
-  end
-  return url
-end
-
-local function get_repo_issues(repo, params)
-  params = params or {}
-  local query_params = {
-    state = params.state or "open",
-    per_page = params.per_page or 50,
-    filter = params.filter,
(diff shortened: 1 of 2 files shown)


lua/octo/completion.lua
```diff
@@ -48,38 +49,51 @@ local function get_repo_issues(repo, params)
...
     local line = api.nvim_get_current_line()
     local pos = vim.fn.col(".")
-    local i, j = 0
+    local i, j = 0, 0
     while true do
       i, j = string.find(line, "#(%d*)", i + 1)
       if i == nil then
-        break
+        i, j = 0, 0
+        i, j = string.find(line, "@(%w*)", i + 1)
```
PatOConnor43, line 62: I couldn't figure out how to get lua's find function to let me do the first regex or the second. This is ugly but seemed to do what I wanted it to do. I'm very open to better ideas here :smile:

lua/octo/init.lua
```diff
@@ -621,6 +652,8 @@ function M.create_issue_buffer(issue, repo, create_buffer)
     200
   )
 
+  async_fetch_taggable_users(bufnr, repo, issue.user.login)
```
PatOConnor43, line 655: It might be cleaner to just not pass `issue.user.login` into this function since I am already setting that value as the default for `taggable_users` [here](https://github.com/pwntester/octo.nvim/pull/48/files#diff-551b860db1f227d6d4d20dfc8a8a09669a68b474e514e85baf9f22c2b4a85e04R474). Then inside the function I can get that variable and append to it. Thoughts?

lua/octo/completion.lua
pwntester: yep, I dont think thats possible with lua patterns, so the workaround is :+1:

lua/octo/init.lua
pwntester: I think thats a better approach, better get `taggable_users` from the buf vars and add new users to it.

lua/octo/completion.lua
```diff
@@ -11,12 +11,13 @@ local M = {}
 local function get_url(url, params)
   url = url .. "?foo=bar"
   for k, v in pairs(params) do
-    url = format("%s\\&%s=%s", url, k, v)
```
pwntester, line 14: dont remember why I was escaping the `&`, probably because before using `gh` I was using a custom `curl` wrapper. thanks for fixing it!

lua/octo/completion.lua
```diff
@@ -48,38 +49,51 @@ local function get_repo_issues(repo, params)
 end
 
 function M.issue_complete(findstart, base)
-  -- :he complete-functions
+  -- the complete-functions
```
pwntester, line 52: :) that was vim `:help` command, but its ok wither way :)

lua/octo/init.lua
```diff
@@ -408,6 +409,32 @@ function M.load_issue()
   )
 end
 
+-- This function accumulates all the taggable users into a single list that
+-- gets set as a buffer variable `taggable_users`. If this list of users
+-- is needed syncronously, this function will need to be refactored.
+-- The list of taggable users should contain:
+--   - The PR author
+--   - The authors of all the existing comments
+--   - The contributors of the repo
```
pwntester, line 418: thanks for documenting the function, that is something I need to do now that people may contribute to the plugin

lua/octo/init.lua
```diff
@@ -444,9 +471,13 @@ function M.create_issue_buffer(issue, repo, create_buffer)
...
   api.nvim_buf_set_var(bufnr, "assignees", issue.assignees)
   api.nvim_buf_set_var(bufnr, "milestone", issue.milestone)
+  api.nvim_buf_set_var(bufnr, "taggable_users", {issue.user.login})
 
   -- local mappings
-  local mapping_opts = {script = true, silent = true}
+  local mapping_opts = {script = true, silent = true, noremap = true}
+
+  api.nvim_buf_set_keymap(bufnr, "i", "@", "@<C-x><C-o>", mapping_opts)
+  api.nvim_buf_set_keymap(bufnr, "i", "#", "#<C-x><C-o>", mapping_opts)
```
pwntester, line 480: that is a great idea, thanks!

lua/octo/completion.lua
```diff
@@ -48,38 +49,51 @@ local function get_repo_issues(repo, params)
...
-        return i
+        -- I think subtracting 1 is necessary to include the first character
+        -- since lua is 1 indexed
+        return i - 1
       end
     end
     return -2
   elseif findstart == 0 then
     local repo = api.nvim_buf_get_var(0, "repo")
     local issues = get_repo_issues(repo)
```
pwntester, line 76: move this line down to the `if vim.startswith(base, "#")` block to avoid loading issues for `@` completion

lua/octo/completion.lua
```diff
@@ -48,38 +49,51 @@ local function get_repo_issues(repo, params)
...
-            abbr = format("#%d", i.number),
-            menu = i.title
-          }
-        )
+    if vim.startswith(base, "@") then
+      local users = api.nvim_buf_get_var(0, "taggable_users") or {}
+      for _, user in pairs(users) do table.insert(entries, {word = format("@%s", user), abbr = user}) end
+    else if vim.startswith(base, "#") then
+        for _, i in ipairs(issues) do
+          if vim.startswith(tostring(i.number), base) then
```
pwntester, line 83:
```suggestion
          if vim.startswith('#'..tostring(i.number), base) then
```

This seems to be necessary now

lua/octo/completion.lua
PatOConnor43: Ah! I always just use `:h` so I didn't recognize it. I'll change it to `:help complete-functions`

pwntester (changes requested):
Thanks for this awesome PR! I suggested some minor changes.
Also, can you please format the code with `luafmt -i 2` to keep code consistency?

PatOConnor43:
Thanks, I can these fixed real quick :+1:

lua/octo/init.lua
PatOConnor43: `#` being a common character in markdown is a good point. I think there's a couple options:
  - Refactor fetching issues to happen in the background and attach to the buffer like users
  - Keep the fetching as-is but also add an `nvim_buf_(get/set)_var` in the completion function that only fetches if the value is nil
  - Remove these mappings so the user has to purposely try to fetch issues/users
  
I'm probably more of a fan of fetching the issues like the users and attaching them to the buffer, but I'll defer to you.

pwntester: That looks good to me and solve the cache problem, so Im ok with it

lua/octo/completion.lua
PatOConnor43: This should be resolved [here](https://github.com/pwntester/octo.nvim/pull/48/commits/a3dba031fc917d4318d87fd27a77ad5a94b31fbf)

lua/octo/completion.lua
PatOConnor43: This should be resolved [here](https://github.com/pwntester/octo.nvim/pull/48/commits/a3dba031fc917d4318d87fd27a77ad5a94b31fbf)

lua/octo/completion.lua
PatOConnor43: This should be resolved [here](https://github.com/pwntester/octo.nvim/pull/48/commits/f6771a23b5f6126baaba6a7e7312daaa14127e7b)

lua/octo/init.lua
PatOConnor43: This should be resolved [here](https://github.com/pwntester/octo.nvim/pull/48/commits/f6771a23b5f6126baaba6a7e7312daaa14127e7b)

lua/octo/init.lua
PatOConnor43: I'll get started on that now :+1:

pwntester: Maybe use a global dict instad of a buffer var so it can be reused by other issue buffers from the same repo

PatOConnor43: I think that's a possibility but I wonder if that's too wide. The issues that you can auto-complete will be defined by the time you open the first issue buffer. So if you want to complete issues that are created later you would need to restart nvim/manually clear that global. Unless we also introduce some cache invalidation?

It may also be worth mentioning that I think a user should be able to make 5000 requests per hour with an api token if you're worried about that.
https://docs.github.com/en/free-pro-team@latest/developers/apps/rate-limits-for-github-apps#normal-user-to-server-rate-limits

pwntester: Ok,that makes sense, lets make it a buf var then

lua/octo/init.lua
```diff
@@ -408,6 +409,48 @@ function M.load_issue()
...
+
+-- This function fetches the issues in the repo so they can be used for
+-- completion.
+local function async_fetch_issues(bufnr, repo)
+  gh.run(
+    {
+      args = {"api", format(format("repos/%s/issues", repo))},
+      cb = function(response)
+        local resp = json.parse(response)
+        api.nvim_buf_set_var(bufnr, "issues", resp)
```
PatOConnor43, line 448: I'm putting the whole issue response in here which we definitely don't need. Would you like me to build smaller tables, similar to what happens with the `comments_metadata`? We probably actually only need title and number.

PatOConnor43:
@pwntester I think this is ready for review again when you have time.

lua/octo/init.lua
PatOConnor43: I went ahead and just made that change since there was so much that wasn't needed :shrug:

pwntester:
Awesome, thanks a lot!

PR #57, opened 2021-01-04
make checks popup look nicer

lukas-reineke:
format the checks output as nice table

Screen Capture_select-area_20210104201456


pwntester:
Thanks!!

PR #58, opened 2021-01-05
Add command completion

lukas-reineke:
command line completion for `Octo` command

For now just the basic commands.
This could be made smarter to also autocomplete issue number after `Octo issue edit` etc

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -695,4 +695,30 @@ function M.issue_interactive_action(action, kind)
   M.issue_action(action, kind, value)
 end
 
+function M.command_complete(args)
+  local command_keys = vim.tbl_keys(commands)
+  local argLead, cmdLine, _ = unpack(args)
+  local parts = vim.split(cmdLine, " ")
+
+  local get_options = function(options)
+    local valid_options = {}
+    for _, option in pairs(options) do
+      if string.sub(option, 1, #argLead) == argLead then
+        table.insert(valid_options, option)
+      end
+    end
+    return valid_options
+  end
+
+  if #parts == 2 then
+    return get_options(command_keys)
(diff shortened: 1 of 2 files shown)


lua/octo/commands.lua
```diff
@@ -695,4 +695,31 @@ function M.issue_interactive_action(action, kind)
...
+        table.insert(valid_options, option)
+      end
+    end
+    return valid_options
+  end
+
+  if #parts == 2 then
+    return get_options(command_keys)
+  end
+  if #parts == 3 then
```
pwntester, line 716: elseif?

lua/octo/commands.lua
```diff
@@ -695,4 +695,31 @@ function M.issue_interactive_action(action, kind)
   M.issue_action(action, kind, value)
 end
 
+function M.command_complete(args)
```
pwntester, line 698: not sure if we should move it to `completion.lua` for tidiness. Im ok either way

pwntester (changes requested):

lua/octo/commands.lua
lukas-reineke: Yeah I thought about doing that first, but then `commands` need to be exported. And having that contained in 1 file is kinda nice.

But I'm also fine either way.

lua/octo/commands.lua
lukas-reineke: `#parts == 2` returns always. But you're right, more explicit to use `elseif`. Pushed a fix

lua/octo/commands.lua
pwntester: ok, lets leave it as is. thanks!

pwntester:
Thanks!!

PR #68, opened 2021-01-14
search

pwntester:


lua/octo/menu.lua
```diff
@@ -731,4 +731,60 @@ function M.changed_files()
...
+              elseif output then
+                local resp = json.parse(output)
+                --print(#resp.data.search.nodes)
+                for _, issue in ipairs(resp.data.search.nodes) do
+                  process_result(gen_from_issue(4)(issue))
+                end
+                process_complete()
+              end
+            end
+          }
```
pwntester, line 776: It seems like there is a debouncing problem and the finder ends up accumulating the results from previous invocations (partial prompts)

@tjdevries @tami5 @conni2461 do you see anything wrong in the custom finder? any ways to improve it and discard results from partial prompts?

tami5:
@pwntester it would be nice if pickers in menu can be mapped to telescope/__extension/octo.lua

so we can do Telescope octo issues.

accept a pr?

pwntester:
sure thing @tami5 thanks!

Issue #118, opened 2021-03-03
Support saving PR review comments


entropitor:
Given github supports drafting Pull Requests through the API, it would be nice if the review comments are not saved in lua state but on the github side. This would allow for your editor to crash, pick off a PR you started online or vice versa

`addPullRequestReview` to create one (event is optional, not setting it should not publish the review yet)
`submitPullRequestReview` to finally submit a PR review

pwntester:
> event is optional, not setting it should not publish the review yet

Great, didnt know that! This looks great to me although @weilbith wanted to work offline in the train :)

We can send `addPullRequestReview` in `Octo review start` (with empty draft `comments`). 
Then add new comments with `addPullRequestReviewComment` as user adds new comments. 
And finally `submitPullRequestReview` on `Octo review submit`

We would also need a new command (`Octo review pending`?) to retrieve "pending" reviews

weilbith:
Wow. This sounds pretty amazing. I'm curious how this looks like at the GitHub side. Does that mean there are objects stored that are hidden in the UI if they don't have specific property set that they are finished or you are the author? :thinking:

One question would be you would handle the case that you don't have a internet connection but add a review comment. @pwntester if I understand your approach correctly this would not work anymore right? Or is there caching for objects that have not been synced yet? What would be then the trigger to sync them once internet is back, without having to finish the whole review?

weilbith:
Maybe we could introduce a queue of objects that need to be pushed? The Queue entries could be extend with an identifier which would allow to replace them, if the object has changed again before it got handled by the queue. This queue could then handle the updates depending on if there is a network connection or not. And then for commands like to finish a review we have to flush the queue synchronously. So such an action which requires flushing would never succeed if there is no network connection and before all preceding data was pushed.

pwntester:
> I'm curious how this looks like at the GitHub site

I think they appear with a yellow PENDING bubble

> Or is there caching for objects that have not been synced yet? 

We still need the lua state to draw the sign column, show the popup (when implemented) etc. So I would just add a "not synched" property to the stored comment table if `addPullRequestReviewComment` fails to know if I need to send it in the final `submitPullRequestReview`

> What would be then the trigger to sync them once internet is back, without having to finish the whole review?

Thats a different approach, but I think its out of scope for now. TBH I would remove offline support altogether

pwntester:
[...]

PR #120, opened 2021-03-03
Synched review comments

pwntester:
Addresses #118 

- [x] save comments to GH
- [x] implement `Octo review  resume`
- [x] implement `Octo review discard`
- [x] on resume: fetch review comments and populate review_comments table
- [x] jump to comment in `Octo review comments`
- [x] fix comment signs

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,11 @@ There is only an `Octo <object> <action> [arguments]` command:
 | | delete | |
 | | move | |
 | review| start| Start a new review |
-| | comments| View in-progress review comments |
 | | submit| Submit the review |
-| | threads | View other people review threads (comment+replies)|
+| | resume| Edit a pending review for current PR |
+| | discard| Deletes a pending review for current PR if any |
+| | comments| View pending review comments |
+| | threads | View all review threads (comment+replies)|
 
 * If repo is not provided, it will be derived from `<cwd>/.git/config`.
 
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -923,8 +923,8 @@ function M.initiate_review(repo, number, pr)
               pull_request_repo = repo,
               pull_request_number = number,
               pull_request_id = pr.id,
-              baseRefSHA = pr.baseRefSHA,
-              headRefSHA = pr.headRefSHA
+              baseRefOid = pr.baseRefOid,
+              headRefOid = pr.headRefOid
             }
           )
         end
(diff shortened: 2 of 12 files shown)


weilbith:
How can I resume a comment in a line I have already added one? I just tried it with `<space>ca` on the exact same line, but this always gives me an empty comment window.

pwntester:
try using `Octo review comments` and then `<c-e>`, or `Octo comment edit` on the line of the comment you want to edit.

Yep, this needs better documentation

weilbith:
Okay so is it a use-case to have multiple comments to the same line of code? Or could we just edit an existing comment if it does exist already? :thinking: 
But I guess it is worth to keep this as a feature. Is there already a mapping for `Octo comment edit`? Like `<space>ce`?

pwntester:
> Okay so is it a use-case to have multiple comments to the same line of code? Or could we just edit an existing comment if it does exist already? 🤔

Not sure. I don't see a reason to add several comments in the same line, but I guess some people will find valid uses cases. This may not be supported currently as the key for `reviews._review_comments` is based on the comment's side, path and lines. So two comments on the same side/path/lines will override the previous one 🤔 

> But I guess it is worth to keep this as a feature. Is there already a mapping for Octo comment edit? Like <space>ce?

No, but it should be. I will add it

weilbith:
> So two comments on the same side/path/lines will override the previous one thinking

You lost me. Why?

>    But I guess it is worth to keep this as a feature. Is there already a mapping for Octo comment edit? Like ce?
 No, but it should be. I will add it
 
 Which reminds me that it is quite hard to add custom mappings. For the issue/PR buffer there is a special filetype and the user can define an `after/ftplugin/...` for it. Wait a sec. This reminds me that the filetype is `octo_issue`. The `ftplugin` syntax for Vim works like that it searches for `ftplugin/{filetype}_*.vim` (and in the `after/ftplugin/{filetype}_*.vim`). So you can have multiple files for the same filetype and postfix them with like `java_mappings.vim` and `java_options.vim`. Or you can define a directory like `ftplugin/{filetype}/*.vim`. But with `octo_issue` this does not work. I had this issue already once back then with another plugin. We want to fix that I think.

pwntester:
> You lost me. Why?

Comments are currently stored in a table using the comment bufname as the key (eg: `octo://pwntester/octo.nvim/pull/94/comment/RIGHT/README.md:128.128`) which means that two comments on the same side/path/startLine/line will collide and the last one will override the first one. I will fix this by using the GitHub id as the key

Issue #99, opened 2021-02-23
Feature request - add hover window for issue/PR numbers on CursorHold


weilbith:
Taken from [here].
The title basically states it. But in the web UI of GitHub you can hover over issue and PR numbers (plus users, also this isn't that interesting) which will show a summary of this object. This is incredibly useful when you read a issue or PR and see such a reference. You usually don't want to interrupt your flow and searching this issue and open it, than search the old issue and scroll to where you left earlier.

weilbith:
Extension: hover over reactions lists users

pwntester:
Been implemented in #129

Some screenshots for comments and feedback:

<img width="938" alt="image" src="https://user-images.githubusercontent.com/125701/110612300-579b4780-8190-11eb-8dd6-4181c0d84cde.png">

<img width="914" alt="image" src="https://user-images.githubusercontent.com/125701/110612318-5cf89200-8190-11eb-9b00-31b69b42a542.png">

<img width="884" alt="image" src="https://user-images.githubusercontent.com/125701/110612343-62ee7300-8190-11eb-812f-d2ca93078efa.png">

<img width="1051" alt="image" src="https://user-images.githubusercontent.com/125701/110612438-7ac5f700-8190-11eb-9bfb-3c132b62fd12.png">

pwntester:
sample query for comment reactions:

```
{
  node(id: "%s") {
    ... on PullRequestReviewComment {
      reactionGroups {
        content
        users(last: 100) {
          nodes {
            login
          }
        }
      }
    }
    ... on PullRequestReview {
      reactionGroups {
        content
        users(last: 100) {
          nodes {
            login
          }
        }
      }
    }
    ... on IssueComment {
      reactionGroups {
        content
        users(last: 100) {
          nodes {
            login
          }
        }
      }
    }
  }
}
```

pwntester:
sample query for user information:

```
{
  user(login: "%s") {
    login
    bio
    company
    followers(first: 1) {
      totalCount
    }
    following(first: 1) {
      totalCount
    }
    hovercard {
      contexts {
        message
      }
    }
    hasSponsorsListing
    isEmployee
    isViewer
    location
    organizations(last: 5) {
      nodes {
        name
      }
    }
    name
    status {
      emoji
      message
    }
    twitterUsername
    websiteUrl
  }
}
```

weilbith:
Reactions are virtual text, aren't they? This means the user can't move the cursor on them. Or do I miss something?

[...]

PR #129, opened 2021-03-10
feature/popup on issue hover

pwntester:
Addresses #99 

- Refactors popup window code into `window.lua`
- Get rid of `popup.nvim` since its not flexible and highly WIP
- Implement hover popups for issues and PRs

### DIFF
--- a/lua/octo/init.lua
+++ b/lua/octo/init.lua
@@ -804,80 +804,19 @@ function M.show_summary()
           api.nvim_err_writeln(stderr)
         elseif output then
           local resp = json.parse(output)
-          local max_length = 80
           local issue = resp.data.repository.issueOrPullRequest
-            local popup_bufnr = api.nvim_create_buf(false, true)
-            local chunks = {}
-            table.insert(chunks, {
-              {repo.." on "},
-              {util.format_date(issue.createdAt), "OctoNvimDetailsValue"}
-            })
-            table.insert(chunks, {
-              {"⊙", "OctoNvimBubbleRed"},
-              {" "..issue.title.." ", "OctoNvimDetailsLabel"},
-              {"#"..issue.number.." ", "OctoNvimDetailsLabel"}
-            })
-            table.insert(chunks, {{""}})
-            table.insert(chunks, {
-              {string.sub(issue.body, 1, max_length - 4 - 2).."…"}
-            })
-            table.insert(chunks, {{""}})
-            if #issue.labels.nodes > 0 then
-              local labels = {}
-              for _, label in ipairs(issue.labels.nodes) do
-                local label_bubble = bubbles.make_label_bubble(
-                  label.name,
-                  label.color,
-                  { margin_width = 1 }
-                )
-                vim.list_extend(labels, label_bubble)
-              end
-              table.insert(chunks, labels)
-              table.insert(chunks, {{""}})
-            end
-            table.insert(chunks, {
-              {vim.g.octo_icon_user or " "},
-              {issue.author.login}
-            })
-            for i=1,#chunks do
-              writers.write_block({""}, {bufnr = popup_bufnr, line = i})
-            end
-            for i=1,#chunks do
-              writers.write_virtual_text(popup_bufnr, constants.OCTO_DETAILS_VT_NS, i-1, chunks[i])
-            end
-            window.create_popup({
-              bufnr = popup_bufnr,
-              width = max_length,
-              height = 2 + #chunks
-            })
+          local popup_bufnr = api.nvim_create_buf(false, true)
+          local max_length = 80
+          local lines = writers.write_issue_summary(popup_bufnr, issue, {max_length = max_length})
+          window.create_popup({
+            bufnr = popup_bufnr,
+            width = max_length,
+            height = 2 + lines
+          })
         end
       end
     }
   )
 end
-
---[[
-pwntester/octo.nvim on Feb 28
-
-<closed icon> Feature request - pop-up window for comments during a review #107
-
-If I have added a comment to a line of code during a review, I can see the highlight …
-
-(UI) (enhancement)
-<user icon> pwntester You commented
-]]--
-
-
---[[
-pwntester/octo.nvim on Mar 8
-
-popup comments #122
-
-Show review comments on popup windows
-
-[master] ← [popup_comments]
-
-<user icon> pwntester You opened
-]]--
 
 return M
--- a/lua/octo/init.lua
+++ b/lua/octo/init.lua
@@ -5,6 +5,7 @@ local util = require "octo.util"
 local graphql = require "octo.graphql"
 local writers = require "octo.writers"
 local folds = require "octo.folds"
+local window = require "octo.window"
 local vim = vim
 local api = vim.api
 local format = string.format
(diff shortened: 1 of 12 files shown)


lua/octo/init.lua
```diff
@@ -761,4 +762,66 @@ function M.apply_buffer_mappings(bufnr, kind)
...
+  local _, repo2 = pcall(api.nvim_buf_get_var, 0, "repo")
+  if not repo2 then
+    return
+  end
+
+  local cursor = api.nvim_win_get_cursor(0)
+  local current_line = vim.fn.getline(".")
+  local different_repo_issue_pattern = "%s([^ /]+/[^ #]+)#(%d+)%s"
+  local same_repo_issue_pattern = "%s#(%d+)%s"
+  local start_col, end_col, repo, number = current_line:find(different_repo_issue_pattern)
```
weilbith: This looks like some logic that is worth to get exported into its own function with a proper name. This reduces the number of abstraction levels of this function and improves the readability. Also this could/should become re-usable (e.g. for hover user names?).

lua/octo/init.lua
```diff
@@ -761,4 +762,66 @@ function M.apply_buffer_mappings(bufnr, kind)
...
+
+  if start_col and end_col and start_col <= cursor[2] and cursor[2] <= end_col then
```
weilbith: Reds complicated. What does this if-statement express? Sometimes it is worth to create local variable with a descriptive name and assign the boolean expression to it. That supports the reader to understand whats going on here.

Despite I don't know what you are doing here: must it be so complicated? Could you provide background information? In my head I have a simple structure like: get word under cursor, check if matches the pattern of an issue reference, fetch the issue and open the preview. All other details should be handled in smaller functions. This concept might be wrong because I don't know the details or you just have an alternative/better approach. But unfortunately I have no chance to understand it. :slightly_frowning_face:

lua/octo/init.lua
```diff
@@ -761,4 +762,66 @@ function M.apply_buffer_mappings(bufnr, kind)
...
+
+popup comments #122
+
+Show review comments on popup windows
+
+[master] ← [popup_comments]
+
+<user icon> pwntester You opened
+]]--
+
```
weilbith: What are these folds for? :confused:

plugin/octo.vim
```diff
@@ -57,6 +57,7 @@ au BufEnter octo://* call s:configure_octo_buffer()
 au BufReadCmd octo://* lua require'octo'.load_buffer()
 au BufWriteCmd octo://* lua require'octo'.save_buffer()
 au CursorHold octo://* lua require'octo.reviews'.show_comment()
+au CursorHold octo://* lua require'octo'.show_summary()
```
weilbith, line 60: Nitpick: I think function names should express what you do. In this case it should first check if there is something to preview and then show it. So I would suggest to extend the name with something like `attempt_to_show_summary` or `eventually_show_summary` or `show_summary_if_cursor_on_issue`. I guess you get the point. Evaluate yourself if you like this idea. :wink:

lua/octo/window.lua
```diff
@@ -25,78 +24,84 @@ function M.create_centered_float(opts)
       max_line = math.max(#line, max_line)
     end
   end
-
   -- calculate window height/width
   local width, height
   if max_line > 0 then
     -- pre-defined content
-    width = math.min(vim_width * 0.9, max_line)
+    width = math.min(vim_width * 0.9, max_line + 4)
```
weilbith, line 31: Where do these "random" numbers from? It will be hard to come back to this code in 2 years and remember what they where for. This either results into the scenario that code remains untouched because developers are afraid to touch it (especially if there are no tests for it) or the developer just deletes it, eventually breaks something which goes undetected (no tests) or he spends many time in research what this was originally for.
Either try to express this number or the math expression as code or (way less preferable) as comment.

lua/octo/window.lua
```diff
@@ -25,78 +24,84 @@ function M.create_centered_float(opts)
...
-
   -- calculate offsets
-  local x_offset = (vim_width - width) / 2
-  local y_offset = (vim_height - height) / 2
-  local header_winid
-  local header_win_opts = {border = {}}
+  local x_offset = math.floor((vim_width - width) / 2)
+  local y_offset = math.floor((vim_height - height) / 2)
+
+  -- outer win (header + border)
```
weilbith, line 45: Comment are meant to express important information about code that the code itself can't express itself. Else comments are a source of many issues for code maintenance (c.f. Robert C. Martin - Clean Code). What does this comment state what the code itself doesn't tell the reader already. If it is about framing code blocks which belong together, than this is strong hint it should be exported into its own function because it is on another abstraction level as the parent function.

lua/octo/init.lua
```diff
@@ -772,32 +772,82 @@ function M.show_summary()
...
-        height = 10
-      })
+      repo = repo2
+      number = number2
     end
   end
+  if not repo or not number then
+    return
+  end
+  local owner = vim.split(repo, "/")[1]
```
weilbith, line 807: This function is waaaay too long. It doesn't fit on a single screen (bad measurement, but a good old one). If I want to get a grasp of how the preview stuff works, I have to fight through all the details from the very top to the bottom. Usually you have forgotten where you came from while you are stuck in the low level details of the function. I would recommend to split it into multiple parts. Another good indicator for if a function is too complex: could you simply imagine the test cases it would need to write for this interface function (assuming you don't use a TDD workflow)?

lua/octo/init.lua
```diff
@@ -768,31 +768,42 @@ function M.show_summary()
...
   end
 
   local repo, number
   local cursor = api.nvim_win_get_cursor(0)
   local current_line = vim.fn.getline(".")
-  local different_repo_issue_pattern = "%s([^ /]+/[^ #]+)#(%d+)%s"
+  local different_repo_issue_pattern = "%s([^/]+/[^#]+)#(%d+)%s"
   local same_repo_issue_pattern = "%s#(%d+)%s"
-  local start_col1, end_col1, repo1, number1 = current_line:find(different_repo_issue_pattern)
+  local url_issue_pattern = "[htps]+://.*github.com/([^/]+/[^/]+)/[pulisue]+/(%d+)"
```
weilbith, line 781: I think you can pull these variables to the outside of the function. This seems to be a "static" pattern which does not alternate for each run of the function. There is no need to re-allocate memory for this variable each time the function runs. It will help to keep the function more clean by removing what is not part of its behavior or to improve readability.

weilbith (commented):
First of all: this feature is so damn nice. Your design looks really great, I like it! :rocket: :tada: 
Feature wise I think I have nothing to add now. The issue should be fulfilled and can be closed after the PR is merged. Later we might want to add hover for users as well?

So I basically have only comments about code quality and slightly architecture. But for the latter I miss a good overview of the structure you added. Certain comments will go into the detail why I fail to do so.
Please don't get my review wrong. My intention was not to bash your work. You did an excellent job in adding this feature! They are (hopefully) constructive feedback on how you can improve in writing code related to exact examples. Many of the comments can be applied to multiple parts of your changes, though I have not put them multiple times (ofc). So I hope you read them, process them and then agree or disagree with it. In this project I have no hard feelings about the code of others. So feel free to ignore any comment. My best wishes are that we can learn something from each over and get better over time for any kind of development.
If you have questions to specific comments feel free to open into the conversation. :upside_down_face:

pwntester:
> Please don't get my review wrong. My intention was not to bash your work. You did an excellent job in adding this feature! They are (hopefully) constructive feedback on how you can improve in writing code related to exact examples. Many of the comments can be applied to multiple parts of your changes, though I have not put them multiple times (ofc). So I hope you read them, process them and then agree or disagree with it. In this project I have no hard feelings about the code of others. So feel free to ignore any comment. My best wishes are that we can learn something from each over and get better over time for any kind of development.

Im totally open to learn (growth mindset) so dont worry to comment/suggest any part of the code. That why I added you as a reviewer after all :smile:

pwntester:
Also, thank you very much for taking the time to contribute, review and share you knowledge!

plugin/octo.vim
pwntester: Im ok to change the name. Will think of a better name

lua/octo/window.lua
pwntester: Yep, they are used to frame code blocks. I will see if it can be split into different functions. Thanks

lua/octo/init.lua
pwntester: Ok, I think I improved enough without splitting it, but may need to do it if I end up adding hovering over user names or other objects

weilbith:
> Im totally open to learn (growth mindset) 

Great attitude! :muscle: 

---

Nice that you applied these changes to quick, iterating over your code to improve it.
Looking forward to get this merged and use it. :hugs:

Issue #127, opened 2021-03-09
Feature request - get all reviewers of a PR, requested or not


weilbith:
> For some yet unknown reason the API always states that there are no reviewers for any PR. 

The API is actually returning the __requested reviewers__, that is the users that the PR author has requested a review from.
These are the ones that in the WebUI appear with a yellow dot.

If we want to have something similar to what the WebUI shows, then we need to retrieve `pullRequestReviews.author.login` and `pullRequestReview.state`. Also, since I user can leave multiple reviews, we need to "merge" them, so we only show the "strongest" one (eg: if a user leaves `CHANGES_REQUESTED` and `COMMENTED`, only `CHANGES_REQUESTED`). I think that is what the WebUI does, but thats probably work for a new PR

_Originally posted by @pwntester in https://github.com/pwntester/octo.nvim/issues/121#issuecomment-793666563_

pwntester:
Fixed in #130

PR #130, opened 2021-03-11
feature/show reviewers

pwntester:
@weilbith, this addresses #127 but its incomplete yet.
Can I have your :eyes: on it? We obviously need to change the review status icons but I would also like to change the colors which is currently not supported by your bubbles implementation
I thought about making `make_bubble` take an array of chunks (text + hls) instead of a string as the first argument. what do you think?
In that case, I think padding should be applied outside of `make_bubble`

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -243,7 +243,7 @@ function M.add_comment()
 
   local comment = {
     createdAt = vim.fn.strftime("%FT%TZ"),
-    author = {login = vim.g.octo_loggedin_user},
+    author = {login = vim.g.octo_viewer},
     body = " ",
     first_comment_id = first_comment_id,
     id = -1,
--- a/lua/octo/init.lua
+++ b/lua/octo/init.lua
@@ -21,7 +21,7 @@ function M.check_login()
       args = {"auth", "status"},
       cb = function(_, err)
         local _, _, name = string.find(err, "Logged in to [^%s]+ as ([^%s]+)")
-        vim.g.octo_loggedin_user = name
+        vim.g.octo_viewer = name
       end
     }
   )
--- a/lua/octo/menu.lua
+++ b/lua/octo/menu.lua
@@ -653,7 +653,7 @@ function M.select_target_project_column(cb)
   end
 
   local owner, name = util.split_repo(repo)
-  local query = graphql("projects_query", owner, name, vim.g.octo_loggedin_user, owner)
+  local query = graphql("projects_query", owner, name, vim.g.octo_viewer, owner)
   gh.run(
     {
       args = {"api", "graphql", "--paginate", "-f", format("query=%s", query)},
--- a/lua/octo/ui/bubbles.lua
+++ b/lua/octo/ui/bubbles.lua
@@ -8,10 +8,10 @@ local highlights = require "octo.highlights"
 -- background. The bubble shape gets especially defined by the outer delimiters.
 -- An examplary usage in this plugin are for label assigned to an issue.
 
-local function make_bubble(content, highlight_group, options)
-  options = options or {}
-  local margin = string.rep(" ", options.margin_width or 0)
-  local padding = string.rep(" ", options.padding_width or 0)
+local function make_bubble(content, highlight_group, opts)
+  opts = opts or {}
+  local margin = string.rep(" ", opts.margin_width or 0)
+  local padding = string.rep(" ", opts.padding_width or 0)
   local body = padding .. content .. padding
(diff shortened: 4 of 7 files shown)


weilbith:
Hmm. I'm mobile atm and did not look at the code yet. But could you provide an example what you want to use the chunks for? It should possible to make the function smarter or actually split it. So we have a basic one which handles the chunk stuff and another one which behaves as the one now, but under the hood uses the chunks implementation.
The padding can be still simply added internally by just manipulating the first and last chunk. I would like to keep it here if possible. But if not, it is also not too much of an issue, just that you can't simply pass the objects properties directly (e.g. the username), but must pad it before.
What confuses me is how you want to use different highlights in a bubble. This could actually be only about foreground colors. This either needs to be documented or somehow checked/guaranteed in the code. How would you maintain the contrast/readability to the independent set background color? But I guess that is neglectable.

pwntester:
Im trying to get something similar to 

image

So I would need to find unicode/devicons icons similar to those and also be able to use the right color for each icon (approved should be green, changes requested should be red, etc.)

Instead of passing a string, I thought about passing something like:

```
{{"user1", "OctoNvimUser"}, {"<approved icon>", "OctoNvimStateApproved"}}
```

We can make the function take a string and do what its doing right now, but if a table is passed, just prepend/append delimiters

weilbith:
Just put the adjustments to the `make_bubble` function aside. How do you want to construct all the highlight groups dynamically and how do you store/configure the colors for the icons?
I mean there is the `OctoNvimUser` highlight group. Now we determine the background color of this group and construct on-the-fly a new group that has no background and the contents background color as foreground. Not we can use that new highlight for the delimiters and the actual body just uses the set configurable highlight group. So this just to summarize that.
Now you want to have the body actually change its font color for certain characters. This means you again have to use the already extracted background color of the configured highlight group. And now you construct again on-the-fly highlights with the same background color, but other foregrounds. If you don't the bubble will "break". We can't assume the highlight groups are perfectly set and work together by having the same background color. This would make customization for the user incredibly hard. I already see the new issues for UI glitches. Also this would make it completely inflexible as the bubble function is very generic. Where do you get the foreground colors for the signs from? How can the user customize them? And this would still mean you can't just adopt `make_bubble` to take highlight chunks, because who is then responsible to construct the highlight groups properly to make it a bubble.

Don't get me wrong. I think there is a solution for this problem. The question is how convenient it will be for the user and for us developers to use/configure. And also how the code will look like.
Do you really think it is that worth to have the signs inside the bubble colored? Seems like a lot of work. I would rather have the whole bubble in a different color to be honest. Thinking about color contrast to make these small icons easy to read inside the bubble sounds cumbersome. 
So my suggestion would to first just add the icons to the bubbles body without anything special about them. Maybe make the icon put at front optional (add parameter to `make_user_bubble` with default true) if it gets to chaotic. And then consider to have alternative highlights for different reviewer states, but for the full bubble. 
But I you really think it is worth the trouble and you want to push the boundaries we will find a good solution for it. Just don't hurt yourself and stress you with such problems. :wink:

weilbith:
Okay. I want to be more productive how to solve the problem in the design how you imagine it.
I think providing a generic solution is a nightmare and will suffer of many issues. So I think the best option is to keep it limited and under control. Therefore I would add a new function which looks something like this `make_reviewer_bubble(name, state_icon, icon_highlight_group)`. Now we can determine the general highlight group for the bubble as for a normal user. Then we determine the foreground color of `icon_highlight_group`. That we use to create a highlight group with the background color of the bubble (we have this for the delimiter already) and the foreground of this specific group. Then we finally can construct all this together to the chunks we need.
This would be a first iteration that works. It won't re-use the `make_bubble` function. Atm I think we gain more trouble to make this function very generic to handle all possible highlight combinations. Though I have something in mind how it could work smoothly. But that could be done in another iteration. First start simple and focused and see if the produced result is usable.
Ofc we need to document these status icon highlight groups that we ignore their background colors. We need to see if we can re-use their `gui` option like being bold...

pwntester:
Thanks for the suggestions, Im open to new designs. We can also:
- do not use bubbles for reviewers so user names go in some constant color and icons in the corresponding color
- use bubbles but do not colorize the icons
- implement the complex solution to have colorized icons in bubbles

I think the first option can be an easy and elegant solution

weilbith:
I totally agree. I'm fine with any solution. But I think for adding this feature first, I would go for the simple solution. And then when this is done there could be another PR to do some fancy visual stuff if needed. But I actually don't think we need this. After all, I actually think there are some solutions to write some good code to achieve what you initially wanted. But I really don't know if this actually works out visually as you maybe would expect.

PR #133, opened 2021-03-14
add Github Enterprise support

skrobul:
This PR adds basic support for Github Enterprise instances.
I briefly tested with the instance I have and the 'PR' and 'Issues" functions work as expected.
Gists do not, due to [underlying bug in GH CLI](https://github.com/cli/cli/issues/1745), there is a relatively easy [workaround available](https://github.com/cli/cli/issues/1745#issuecomment-694377743).

In fact, setting `GH_HOST` environment variable kinda renders this pull request useless because it makes the PR/issue commands work out of the box too so I would like to get your input/preferences on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ The (addition) `viewer` means the user of the plugin or more precisely the user 
 - `g:octo_icon_user`: Icon used to signal user names (default: "")
 - `g:octo_icon_reaction_viewer_hint`: Icon as alternative or to complement the highlighting of reactions by the viewer himself (default: "")
 - `g:octo_snippet_context_lines`: Number of additional lines displayed from the diff-hunk for single-line comments (default: 3)
+- `g:octo_github_hostname`: Hostname to use for non-public, GHE (GitHub Enterprise) instances (default: null).
 
 ## FAQ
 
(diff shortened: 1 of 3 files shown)


pwntester (approved):

pwntester:
Thanks for the contribution. I dont have a strong opinion whether we should rely on using `GH_HOST` externally or handling it internally. I think it does not hurt to handle it within Octo so Im ok merging this PR in.

However, this reminded me that there are a couple of places where we should account for the GHE hostname:
- `util.lua:508`. Parsing URLs to extract repo name and issue number
```
function M.parse_url(url)
  local repo, kind, number = string.match(url, "https://github.com/([^/]+/[^/]+)/([^/]+)/(%d+)")
  if repo and number and kind == "issues" then
    return repo, number, "issue"
  elseif repo and number and kind == "pull" then
    return repo, number, kind
  end
end
```

pwntester:
After refactoring some code, the only URL handling code should be:

[`constants.lua`]:

```
M.URL_ISSUE_PATTERN = "[htps]+://.*github.com/([^/]+/[^/]+)/([pulisue]+)/(%d+)"
```

skrobul:
Thanks for that, I missed the parse_url stuff since all of my repos use SSH. Anyways, I've rebased on top of current master branch and updated `URL_ISSUE_PATTERN` to include hostname setting as the domain is usually in format of `github.company.net` and similar.

pwntester:
Thanks!

PR #137, opened 2021-04-04
feature/review refactor

pwntester:
This PR unifies the UX regarding PR reviews. We move away from using popups to enter comments and instead use the same kind of issue/pr buffer the user is used to.
We also unified `Octo review start|resume` and `Octo review threads`. The later is removed and all threads are now shown in the review diff buffers.
This allow users to see other people review comments while doing their own reviews. It also enables users to add review comments to other people threads

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ There is only an `Octo <object> <action> [arguments]` command:
 | | edit | [repo] <number> |
 | | list | [repo] [key=value]*<br>[Available keys](https://docs.github.com/en/free-pro-team@latest/graphql/reference/input-objects#issuefilters)<br>Mappings:<br>`<CR>`: Edit issue<br>`<C-b>`: Opens issue in web browser |
 | | search | |
-| | reload | |
+| | reload | same as doing `e!`|
 | | browser | |
 | pr | list | [repo] [key=value]<br>[Available keys](https://docs.github.com/en/free-pro-team@latest/graphql/reference/input-objects#issuefilters)<br>Mappings:<br>`<CR>`: Edit PR<br>`<C-b>`: Opens PR in web browser<br>`<C-o>`: Checkout PR |
 | | search | |
@@ -56,7 +56,7 @@ There is only an `Octo <object> <action> [arguments]` command:
 | | merge | [commit\|rebase\|squash] [delete] |
 | | ready| |
 | | checks | |
-| | reload | |
+| | reload | same as doing `e!`|
 | | browser | |
 | gist | list | [repo] [key=value]*<br>[Available keys](https://cli.github.com/manual/gh_gist_list):  `repo`\|`public`\|`secret`<br>Mappings:<br>`<CR>`: Append Gist to buffer<br>`<C-b>`: Opens Gist in web browser |
 | comment | add | |
@@ -84,7 +84,6 @@ There is only an `Octo <object> <action> [arguments]` command:
 | | resume| Edit a pending review for current PR |
 | | discard| Deletes a pending review for current PR if any |
 | | comments| View pending review comments |
-| | threads | View all review threads (comment+replies)|
 
 * If repo is not provided, it will be derived from `<cwd>/.git/config`.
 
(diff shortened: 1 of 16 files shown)


pwntester:
@weilbith I know pull reviews at the commit level are still not supported. But I wanted to implement this first since it drastically reduces the previous code and removes a lot of duplicate code. Now octo buffers are used all around.
There may be too many code changes to review, but I would appreciate a UX/usability review.

weilbith:
Hey, I'm just reviewing this PR from Octo (this comment here is in the browser now). But I'm confused.
So I have the usual split view in review mode with left and right version, plus the quickfix window at the bottom. I now add a new comment to a line in the right version. Doing so (i.e. press `<space>ca`), the left version gets replaced with a buffer to add a new comment. This looks nice and as you said it looks like a issue buffer with the `Thread` topic. Just that the only content is the snippet the comment I can now add. Alright. Comment written. But what do I do now? I safe the buffer. The comment does not show as "dirty" anymore. It still shows state pending (which is correct) and I can see the comment here in the browser. But how do I close this window again? If I just close it, I'm left with the right version only plus the quickfix window. Also if I execute `:cc` to "re-focus" the current quickfix entry this changes nothing. Just when I go to the next quickfix entry (which means the next whole file) I get back to the left-right diff view...
The same works with comments on the left diff view window.

Can you help me?

weilbith:
I regularly get this error when I stop on lines with a comment:

```
Error detected while processing CursorHold Autocommands for "octo://*":
E5108: Error executing lua ...test_site/pack/packer/opt/octo.nvim/lua/octo/reviews.lua:612: Vim:E95: Buffer with this name already exists
```

pwntester:
> Can you help me?

The idea (which seems like its not obvious 😄 ) is that you should move back to the window you come from and then move the cursor.

So basically, cursorhold will show the comments, then you can move to the window with the comment, review it, add comments and when done, come back to the original window and move the cursor

pwntester:
> I regularly get this error when I stop on lines with a comment:
> 
> 
> 
> ```
> 
> Error detected while processing CursorHold Autocommands for "octo://*":
> 
> E5108: Error executing lua ...test_site/pack/packer/opt/octo.nvim/lua/octo/reviews.lua:612: Vim:E95: Buffer with this name already exists
> 
> ```
> 
> 
> 
> This might be a powerful feature if it works?
> 
> 
> 
> PS: I fail to edit comments I already edited. How do I do this now? 🤔 

Seems like for some reason you are creating a new thread buffer when a previous one for the same file/line already exists. Will look into it

weilbith:
> > Can you help me?
> 
> The idea (which seems like its not obvious 😄 ) is that you should move back to the window you come from and then move the cursor.
> 
> So basically, cursorhold will show the comments, then you can move to the window with the comment, review it, add comments and when done, come back to the original window and move the cursor

Haha okay. :laughing:
I tried to follow your adopted steps in the README. ^^
Hmm. But yeah, it is not really obvious or intuitive for me that I have to go back, then this thread closes if I move. Though the idea sounds nice. Can I somehow still have the cursor on a line I commented and just see the diff for it? Do I somehow need to stay in a different line? Not sure how it works in detail (yet, as I did not came so far), but what happens for range comments? Because this could complicate the "put the cursor above the comment to see the diff" for large ranges. But that is a detail.

TL;DR I think you have pretty cool idea. But it definitely needs some tweaks to make the flow obvious/well working.
Why do you still think floating windows are bad for this? I think I forgot your arguments.
But really just this solution with floating windows sounds perfect to me. So you hover over a line with a comment and a small (large enough to show the thread) floating window shows below the last line the comment range relates to (so for single line comments this is just below the cursor). This has the advantage that I still see the full diff view with left **and** right version. I currently have the issue that if I open the comment thread it replaces one of both version. I can't see the other one anymore.
If you add a new comment in a floating windows, it is kinda obvious that you can save and close the buffer/window and just "returns" you to the previous view of left, right + qf. And the flow does not get more complicated. You could still have a single mapping/command to add or edit a comment. When you are on a line without a comment, this opens a the floating window with a fresh thread and put your cursor into the window. If the line already has a comment it automatically gets shown (as your current approach does it) and the same mapping/command just navigates you into this floating window.

Sorry for fighting so often for floating windows. I just think they are a really good solution for this case. They solve issues other approaches have, plus bring additional benefits automatically (e.g. presenting the diff-view while writing comments).

weilbith:
Cool that you work on this. You definitely set my excitement about this plugin on fire again. Sorry that still can't use it in practise. :see_no_evil:

weilbith:
Btw: I solved the lazy loading of this plugin and it's dependencies in Packer. Though this is only a temporally solution as "we" want to improve the general behavior in Packer for such cases. The issue was to lazy load Octo, but first bring Telescope into scope as else the whole Octo plugin fails to load unless you restart NeoVim. Another combination of plugins had a similar issue so I fixed it.

weilbith:
> > I regularly get this error when I stop on lines with a comment:
> > 
> > 
> > 
> > ```
> > 
> > Error detected while processing CursorHold Autocommands for "octo://*":
> > 
> > E5108: Error executing lua ...test_site/pack/packer/opt/octo.nvim/lua/octo/reviews.lua:612: Vim:E95: Buffer with this name already exists
> > 
> > ```
> > 
> > 
> > 
> > This might be a powerful feature if it works?
> > 
> > 
> > 
> > PS: I fail to edit comments I already edited. How do I do this now? 🤔 
> 
> Seems like for some reason you are creating a new thread buffer when a previous one for the same file/line already exists. Will look into it

I think the issue here is simple now I read your explanation. As I did not understood how it works I wrote the buffer and closed the window. I did not delete the buffer. And the autocommand now tries to open/create a new buffer with the same name. But the old one is still loaded and just hidden. Thereby it fails. :shrug:

weilbith:
With all your information I now restarted NeoVim to get a clear buffer list etc and try it again to see how cool it is. I used `Octo review resume` and moved the cursor to one of the lines with a pending comment. Now I get a different error:

```
Error detected while processing CursorHold Autocommands for "octo://*":
E5108: Error executing lua ...test_site/pack/packer/opt/octo.nvim/lua/octo/writers.lua:634: attempt to perform arithmetic on field 'foldcolumn' (a string value)
```

pwntester:
Both 
> ```
> Error detected while processing CursorHold Autocommands for "octo://*":
> E5108: Error executing lua ...test_site/pack/packer/opt/octo.nvim/lua/octo/reviews.lua:612: Vim:E95: Buffer with this name already exists
> ```

and

> ```
> Error detected while processing CursorHold Autocommands for "octo://*":
> E5108: Error executing lua ...test_site/pack/packer/opt/octo.nvim/lua/octo/writers.lua:634: attempt to perform arithmetic on field 'foldcolumn' (a string value)
> ```

should be fixed now

> I think the issue here is simple now I read your explanation. As I did not understood how it works I wrote the buffer and closed the window. I did not delete the buffer. And the autocommand now tries to open/create a new buffer with the same name. But the old one is still loaded and just hidden. Thereby it fails. :shrug:

Yep, I fixed by reusing the buffer if it already exists

> Hmm. But yeah, it is not really obvious or intuitive for me that I have to go back, then this thread closes if I move. Though the idea sounds nice. 

Ive been using this approach and once you know how it works and get used to it, I think its the best way to handle so many things in a TUI

> Can I somehow still have the cursor on a line I commented and just see the diff for it? Do I somehow need to stay in a different line? 

If there are comments, yep, you need to move to a line with no comments. Alternatively, I can replace `cursorhold` with a mapping. But I think it works fine as is

> Not sure how it works in detail (yet, as I did not came so far), but what happens for range comments? Because this could complicate the "put the cursor above the comment to see the diff" for large ranges. But that is a detail.

Yep, comments over large ranges will effectively prevent vieweing the other diff buffer 🤔

> But really just this solution with floating windows sounds perfect to me. So you hover over a line with a comment and a small (large enough to show the thread) floating window shows below the last line the comment range relates to (so for single line comments this is just below the cursor). This has the advantage that I still see the full diff view with left **and** right version. I currently have the issue that if I open the comment thread it replaces one of both version. I can't see the other one anymore.

>Sorry for fighting so often for floating windows. I just think they are a really good solution for this case. They solve issues other approaches have, plus bring additional benefits automatically (e.g. presenting the diff-view while writing comments).

May consider this approach in the future as a configurable alternative to the regular window but I still see focusing in/out of a float win a little clumsy

> Btw: I solved the lazy loading of this plugin and it's dependencies in Packer. Though this is only a temporally solution as "we" want to improve the general behavior in Packer for such cases. The issue was to lazy load Octo, but first bring Telescope into scope as else the whole Octo plugin fails to load unless you restart NeoVim. Another combination of plugins had a similar issue so I fixed it.

Cool, I dont have that issue but we may want to add it to the README

PR #151, opened 2021-05-30
change prompt_prefixes to prompt_titles

### LINKED ISSUE #150: prompt prefix vs title
I noticed that this project uses `prompt_prefix` to describe the context/actions: e.g. `Issue Search> `. What's your opinion on changing them to `prompt_title`s instead? The argument here being that using the titles would be consistent with all the other Telescope's pickers as they change the title instead of the prefix.

rudotriton:

This change brings about some consistency with with all the other Telescope's prompts, by setting the prompt title instead of the prefix. This also preserves users' custom prompt prefixes in case they have set any.

closes #150 

### Describe how to verify it

Running any of the Octo's commands should now show the "action" as a prompt title instead of the prefix.

What Octo's prompts look like:

Screen Shot 2021-05-30

What Telescope's inbuilt prompts look like (with a user-set prompt_prefix):

Screen Shot 2021-05-30

What Octo's prompts would be changed to (with a user-set prompt_prefix):

Screen Shot 2021-05-30

### DIFF
--- a/lua/octo/menu.lua
+++ b/lua/octo/menu.lua
@@ -605,7 +605,7 @@ function M.select_project_card(cb)
     pickers.new(
       opts,
       {
-        prompt_title = "Choose card",
+        prompt_title = "Choose Card",
         finder = finders.new_table {
           results = cards.nodes,
           entry_maker = entry_maker.gen_from_project_card()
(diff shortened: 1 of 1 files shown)


pwntester:
Cool, thanks!

PR #183, opened 2021-07-27
Fix PR checkout from telescope list

morgsmccauley:

Checking out PR from `Octo pr list` does not work. The telescope entry is being interpreted incorrectly, `selection.value` is the PR number rather than a table.

### Describe how you did it
Use the correct number when attempting to checkout a PR

### Describe how to verify it
Tested the change locally


pwntester:
Thanks a lot, I will be changing this function to reuse `commands.checkout_pr` which does not depend on `gh` and is a lo faster. Please test it if you can

pwntester:
I implemented that change on https://github.com/pwntester/octo.nvim/pull/184 but it is now in master

PR #189, opened 2021-08-11
fix: correct function checkout_pr points to

fdschmidt93:

Checking out a PR got refactored in #184 and now the mapping points to a non-existing function in `commands.lua`.

No issue exists but happy to add one post-hoc if desired.

### Describe how you did it

Changing what command `checkout_pr` in `mappings.lua` points to, namely `commands.pr.checkout`

### Describe how to verify it

Try checking out as is and you'll run into an error. This PR is a simple fix for that :)


pwntester:
Thanks!!

PR #192, opened 2021-08-15
🐛 FIX: mappings commands

### LINKED ISSUE #191: attempt to call field 'add_create'
Trying to add a new label with default mapping `<leader>lc` results in the
following error:

```
E5108: Error executing lua ...octo/mappings.lua:57: attempt to call field 'add_create' (a nil value)
```

tami5:
#### Purpose

Try and fix some mappings. closes #191.

#### Details

I'm using stylua so it auto formatted the buffer a bit. I strongly
propose to use stylua and @pwntester if you like, I can set it up in
another pr.

### DIFF
--- a/lua/octo/mappings.lua
+++ b/lua/octo/mappings.lua
@@ -12,172 +12,186 @@ end
 
 M.keypress_event_cbs = {
   close_issue = function()
-    require"octo.commands".change_state('CLOSED')
+    require("octo.commands").change_state "CLOSED"
   end,
   reopen_issue = function()
-    require"octo.commands".change_state('OPEN')
+    require("octo.commands").change_state "OPEN"
   end,
   list_issues = function()
     local bufnr = vim.api.nvim_get_current_buf()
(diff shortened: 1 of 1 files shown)


pwntester:
Thanks @tami5!
That would be nice! Please send the PR to check the format (github action?) and reformat all files.

What plugin do you use to apply stylua?

tami5:
> Thanks @tami5!
> That would be nice! Please send the PR to check the format (github action?) and reformat all files.
> 
> What plugin do you use to apply stylua?

sure thing, I have set it up in a plugin before and it should be a quick and easy pr.

I'm using null_ls, it has some issues here and there but overall working good. 

my config:

```lua
function()
    local null_ls = require "null-ls"
    local b = null_ls.builtins
    null_ls.setup {
      sources = {
        b.formatting.stylua.with {
        args = {
          "--config-path",
          vim.env.HOME .. "/.config/nvim/stylua.toml",
          "-",
        },
      },
    }
  end
```

PR #231, opened 2021-12-29
Fix error recognition for `git remote` command

### LINKED ISSUE #230: Error: could not resolve repo 'nil/nil' after change to using `git remote` command
### Description

After the change in https://github.com/pwntester/octo.nvim/pull/229, running `:Octo issue list` inside my repo gives the error:

```
gh: Could not resolve to a Repository with the name 'nil/nil'.
```

I noticed there is a change of `git config --get` to `git remote get-url`, and I think because the latter prints an error when the remote doesn't exist as opposed to empty, this occurs.

I have `default_remote` set to the default in my config:

```lua
  default_remote = {"upstream", "origin"}; -- order to try remotes
```

The result of running the commands in the terminal:

```
$ git config --get remote.upstream.url
$ git config --get remote.origin.url
https://github.com/farbodsz/my-repo.git
$ git remote get-url upstream
fatal: No such remote 'upstream'
$ git remote get-url origin
https://github.com/farbodsz/my-repo.git
```

Instead of [checking the output string for "error"], perhaps "fatal" would be better.

farbodsz:
Fixes #230 

Checking for the string "error" can be problematic: 

- Firstly, errors can return other strings like "fatal", which will not be caught
- A repository url could contain error (e.g. my-error-checker.git or similar), and this will lead to false errors being caught

Using `/dev/null`, we can check for an empty string ignoring error messages.

I've tested it locally and it works for me.

Caveat: this obviously won't work with Windows machines, but not sure if we're supporting them anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Edit and review GitHub issues and pull requests from the comfort of your favorit
 ## ⚡️Requirements
 
 - Install [GitHub CLI](https://cli.github.com/)
+- Install [plenary.nvim](https://github.com/nvim-lua/plenary.nvim)
 - Install [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim)
 - Install [nvim-web-devicons](https://github.com/kyazdani42/nvim-web-devicons)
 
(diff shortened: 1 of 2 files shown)


pwntester (commented):
Thanks for the PR, perhaps it would be better to use Plenary's jobs to run the command so we control stderr and stdout without pipe redirections

farbodsz:
Thanks Alvaro - yes plenary is a much better idea than the piping.

I've made the changes and updated the README to note the requirement on plenary (even though it's already used across the repo anyway)

pwntester (approved):

pwntester:
Awesome, thanks!

Issue #235, opened 2022-01-10
Allow gh to use GITHUB_TOKEN env var for auth


cwndrws:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

It would be really nice if octo could use the `GITHUB_TOKEN` env var to authenticate. I ran into this while trying to use octo in a codespace. In a codespace, `GITHUB_TOKEN` is set so that tools that consume it ,  `gh` for example, work out of the box. It surprised me to see the, message about having to auth with `gh auth login` since `gh` was working fine with the env var set.

### Describe what you expected to happen

I expect octo to work the same as `gh` in that it allows me to auth with the `GITHUB_TOKEN` env var.

PR #236, opened 2022-01-10
Add gh token to env vars passed to gh to allow for token auth

cwndrws:

Allows the `GITHUB_TOKEN` env var to be sent through to `gh` when making calls out to it. This allows octo to work out of the box in environments where `GITHUB_TOKEN` is automatically set (Codespaces, for example).

Fixes https://github.com/pwntester/octo.nvim/issues/235

### Describe how you did it

I added `GITHUB_TOKEN` to list of environment variables that gets passed to the job that calls out to `gh`

### Describe how to verify it

I ran my version of octo in a codespace where `GITHUB_TOKEN` was set and it worked.


pwntester:
Cool, thanks a lot!

PR #243, opened 2022-01-19
Ensure all files from PR are loaded

pgherveou:

When `gh` returns multiple pages, the plugins break cause the output is not a JSON file but the concatenation of multiple JSON files.

NONE

### Describe how you did it

One way to break it down is to use the `--jq` argument that will return each JSON content on a different page.
 We can then split the returned content and populate the results object with all the PR files

### Describe how to verify it

I guess one could create a repo with a PR that generates multiple diffs to enforce the pagination.
For me it breaks with the following stack trace:

```
Error executing vim.schedule lua callback: Vim:E474: Trailing characters: ... stack traceback:
        [C]: in function 'json_decode'
        .../.config/nvim/plugged/octo.nvim/lua/octo/reviews.lua:149: in function 'cb'
        .../.config/nvim/plugged/octo.nvim/lua/octo/gh.lua:83: in function 'cb'
```

### Special notes for reviews

After applying this patch, It looks like the plugin can also break as it tries to load too many files. (it throws a   pid = "EMFILE: too many open files"). I haven't had time to investigate this part, but it might require follow-up work.

### DIFF
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { 'api', '--paginate', url, '--jq', '.' },
+    args = { "api",  "--paginate", url, "--jq",  "." },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.notify(stderr, 2)
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { "api", "--paginate", url, "--jq",  "." },
+    args = { "api", "--paginate", url, "--jq", "." },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.notify(stderr, 2)
@@ -146,9 +146,8 @@ function Review:initiate()
           deleted = "D",
           renamed = "R",
         }
-
         local results = {}
-        local page_outputs = vim.split(output, '\n')
+        local page_outputs = vim.split(output, "\n")
         for _, text in ipairs(page_outputs) do
           local page_result = vim.fn.json_decode(text)
           for _, result in ipairs(page_result) do
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { "api", "--paginate", url },
+    args = { 'api', '--paginate', url, '--jq', '.' },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.notify(stderr, 2)
@@ -146,7 +146,16 @@ function Review:initiate()
           deleted = "D",
           renamed = "R",
         }
-        local results = vim.fn.json_decode(output)
+
+        local results = {}
+        local page_outputs = vim.split(output, '\n')
+        for _, text in ipairs(page_outputs) do
+          local page_result = vim.fn.json_decode(text)
+          for _, result in ipairs(page_result) do
+            table.insert(results, result)
+          end
+        end
+
         local files = {}
         for _, result in ipairs(results) do
           local entry = FileEntry:new {
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { "api",  "--paginate", url, "--jq",  "." },
+    args = { "api", "--paginate", url, "--jq",  "." },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.notify(stderr, 2)
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -147,7 +147,7 @@ function Review:initiate()
           renamed = "R",
         }
 
-        local results = utils.get_pages(output)
+        local results = utils.get_flatten_pages(output)
         local files = {}
         for _, result in ipairs(results) do
           local entry = FileEntry:new {
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -146,15 +146,8 @@ function Review:initiate()
           deleted = "D",
           renamed = "R",
         }
-        local results = {}
-        local page_outputs = vim.split(output, "\n")
-        for _, text in ipairs(page_outputs) do
-          local page_result = vim.fn.json_decode(text)
-          for _, result in ipairs(page_result) do
-            table.insert(results, result)
-          end
-        end
 
+        local results = utils.get_pages(output)
         local files = {}
         for _, result in ipairs(results) do
           local entry = FileEntry:new {
--- a/lua/octo/reviews/file-panel.lua
+++ b/lua/octo/reviews/file-panel.lua
@@ -303,6 +303,9 @@ function FilePanel:render()
     offset = #s
 
     -- viewer viewed state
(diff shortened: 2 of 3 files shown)


lua/octo/reviews.lua
```diff
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { "api", "--paginate", url },
+    args = { 'api', '--paginate', url, '--jq', '.' },
```
pgherveou, line 138:
```suggestion
    args = { "api",  "--paginate", url, "--jq",  "." },
```

lua/octo/reviews.lua
```diff
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { "api", "--paginate", url },
+    args = { "api",  "--paginate", url, "--jq",  "." },
```
pgherveou, line 138:
```suggestion
    args = { "api", "--paginate", url, "--jq",  "." },
```

pgherveou:
I was able to quickly fix the other errors by running the jobs sequentially.
There might be some better strategy, but here is the diff in case this can help.

There were also some missing nil check in the  file-panel that I needed to add to finally make the plugin work on my PR
```diff
diff --git a/lua/octo/reviews.lua b/lua/octo/reviews.lua
index 0797114..bdaaf5d 100644
--- a/lua/octo/reviews.lua
+++ b/lua/octo/reviews.lua
@@ -135,7 +135,7 @@ function Review:initiate()
 
   local url = string.format("repos/%s/pulls/%d/files", pr.repo, pr.number)
   gh.run {
-    args = { "api", "--paginate", url },
+    args = { "api", "--paginate", url, "--jq", "." },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.notify(stderr, 2)
@@ -146,7 +146,15 @@ function Review:initiate()
           deleted = "D",
           renamed = "R",
         }
-        local results = vim.fn.json_decode(output)
+        local results = {}
+        local page_outputs = vim.split(output, "\n")
+        for _, text in ipairs(page_outputs) do
+          local page_result = vim.fn.json_decode(text)
+          for _, result in ipairs(page_result) do
+            table.insert(results, result)
+          end
+        end
+
         local files = {}
         for _, result in ipairs(results) do
           local entry = FileEntry:new {
diff --git a/lua/octo/reviews/file-panel.lua b/lua/octo/reviews/file-panel.lua
index 951e461..088f95f 100644
--- a/lua/octo/reviews/file-panel.lua
+++ b/lua/octo/reviews/file-panel.lua
@@ -303,6 +303,9 @@ function FilePanel:render()
     offset = #s
 
     -- viewer viewed state
+    if not file.viewed_state then
+      file.viewed_state = "UNVIEWED"
+    end
     local viewerViewedStateIcon = utils.viewed_state_map[file.viewed_state].icon
     local viewerViewedStateHl = utils.viewed_state_map[file.viewed_state].hl
     s = s .. " " .. viewerViewedStateIcon
diff --git a/lua/octo/utils.lua b/lua/octo/utils.lua
index 2467a32..43f4e3e 100644
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -190,22 +190,28 @@ function M.commit_exists(commit, cb)
     :start()
 end
 
+local last_get_file_job = nil
+
 function M.get_file_at_commit(path, commit, cb)
   if not Job then
     return
   end
-  Job
-    :new({
-      enable_recording = true,
-      command = "git",
-      args = { "show", string.format("%s:%s", commit, path) },
-      on_exit = vim.schedule_wrap(function(j_self, _, _)
-        local output = table.concat(j_self:result(), "\n")
-        local stderr = table.concat(j_self:stderr_result(), "\n")
-        cb(vim.split(output, "\n"), vim.split(stderr, "\n"))
-      end),
-    })
-    :start()
+  local job = Job:new {
+    enable_recording = true,
+    command = "git",
+    args = { "show", string.format("%s:%s", commit, path) },
+    on_exit = vim.schedule_wrap(function(j_self, _, _)
+      local output = table.concat(j_self:result(), "\n")
+      local stderr = table.concat(j_self:stderr_result(), "\n")
+      cb(vim.split(output, "\n"), vim.split(stderr, "\n"))
+    end),
+  }
+  if last_get_file_job ~= nil then
+    last_get_file_job:and_then(job)
+  else
+    job:start()
+  end
+  last_get_file_job = job
 end
 
 function M.in_pr_repo()
```

pwntester:
Thanks! Didnt know about the `--jq` option. I already have a utility function to aggregate graphql responses in `utils.aggregate_pages`. This function uses `get_pages` to get the different responses. It seems like we could use the `--jq` option to change the way that `get_pages` works. Your approach is definitively cleaner than mine 😄

pwntester:
> I was able to quickly fix the other errors by running the jobs sequentially. There might be some better strategy, but here is the diff in case this can help.
> 
> There were also some missing nil check in the file-panel that I needed to add to finally make the plugin work on my PR

Can you please add them to the PR?

pwntester:
> Thanks! Didnt know about the `--jq` option. I already have a utility function to aggregate graphql responses in `utils.aggregate_pages`. This function uses `get_pages` to get the different responses. It seems like we could use the `--jq` option to change the way that `get_pages` works. Your approach is definitively cleaner than mine 😄

Just implemented this on https://github.com/pwntester/octo.nvim/pull/244

lua/octo/reviews.lua
```diff
@@ -146,7 +146,8 @@ function Review:initiate()
           deleted = "D",
           renamed = "R",
         }
-        local results = vim.fn.json_decode(output)
+
```
pgherveou, line 149:
```suggestion
```

lua/octo/reviews/file-panel.lua
```diff
@@ -303,6 +303,9 @@ function FilePanel:render()
     offset = #s
 
     -- viewer viewed state
+    if not file.viewed_state then
```
pgherveou, line 306: not sure why, but I had to add this check to make it work on my PR

lua/octo/utils.lua
```diff
@@ -190,22 +190,28 @@ function M.commit_exists(commit, cb)
...
+      local stderr = table.concat(j_self:stderr_result(), "\n")
+      cb(vim.split(output, "\n"), vim.split(stderr, "\n"))
+    end),
+  }
+  if last_get_file_job ~= nil then
+    last_get_file_job:and_then(job)
+  else
+    job:start()
+  end
+  last_get_file_job = job
```
pgherveou, line 214: this should, make each job run sequentially, ideally there is some kind of concurrency parameter on Job, so that we can run more than one at a time...

pgherveou:
added the other diff and rebase against the main branch. 
⚠️ These changes have just been tested against a single PR, and not sure whether or not they are tested in this repo

lua/octo/utils.lua
pwntester: Nice, have you noticed any performance impact?

pwntester (approved):

pgherveou:
actually the get_pages from util is missing one inner loop to flatten the results from the page
see https://github.com/pwntester/octo.nvim/pull/243/commits/6c295647d5f480cb05de287b2e3ccf9d5d7cf2c2
I guess you dont need that with aggregate_pages cause you are dealing with json objects at the root and not json arrays?

pwntester:
> actually the get_pages from util is missing one inner loop to flatten the results from the page see [6c29564](https://github.com/pwntester/octo.nvim/commit/6c295647d5f480cb05de287b2e3ccf9d5d7cf2c2) I guess you dont need that with aggregate_pages cause you are dealing with json objects at the root and not json arrays?

Yep, the `get_pages` should be agnostic of what its dealing with. `aggregate_pages` works with json objects and cannot be used for json arrays, but it may make sense to add a similar function for aggregating arrays.

lua/octo/utils.lua
pgherveou: I have a PR with a hundred of file changes, this is quite fast to run, I haven't done any measures, but it should not be  a noticeable changes for most PR that usually have less changes than that

pgherveou:
done, feel free to merge / amend when you want.
and thanks again for making this plugin available 🙏

PR #250, opened 2022-01-22
Add action to list all available Octo actions

### LINKED ISSUE #163: [Feature request] Global Telescope menu
## Issue Description

Type: *feature request*

### Describe

First, thanks for an amazing plugins. Don't know whether it was asked before, but here is my proposal.

I want octo.nvim to have a global Telescope menu (either bound to `:Octo` or `:Octo commands`), from which users can access other commands. This will make key bindings for Octo easier (literally just need 1 for all functionalities).

The results window will prompt for `repo card review ... issue`, then next after choosing `issue` it will prompt for `list create ... reopen` (much like the current command auto-suggestion, but with Telescope interface). The preview window can describe what the currently selected command does, for example.

ben-davis:
Adds an action that lists all available Octo actions and runs them when selected.

Fixes #163. 

### Describe how you did it
1. Added a new top-level `Octo actions` action. This required adding support for actions without an object.
2. This iterates through each object in `commands.lua` and flattens them into an array that's then passed to the provider.
3. The provider renders each action with two columns: `object action`, e.g. `pr list`.

### Describe how to verify it
Run `Octo actions`.

### Special notes for reviews
This currently doesn't handle actions that require arguments like `issue create`. I see two options for that:
1. When building the `flattened_actions` array, we inspect the given action to determine if it requires an argument and then prompt the user for each argument.
2. Adjust existing actions that require arguments so they can be called without the argument, and then prompt the user if the args aren't given.

@pwntester what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ If no command is passed, the argument to `Octo` is treated as a URL from where a
 | | resume| Edit a pending review for current PR |
 | | discard| Deletes a pending review for current PR if any |
 | | comments| View pending review comments |
+| octo | actions | Lists all available Octo actions|
 
 0. `[repo]`: If repo is not provided, it will be derived from `<cwd>/.git/config`.
 
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ If no command is passed, the argument to `Octo` is treated as a URL from where a
 | | resume| Edit a pending review for current PR |
 | | discard| Deletes a pending review for current PR if any |
 | | comments| View pending review comments |
-| octo | actions | Lists all available Octo actions|
+| actions |  | Lists all available Octo actions|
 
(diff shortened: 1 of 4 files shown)


pwntester:
Thanks, this fixes #163
Im wondering if we should bind this to `:Octo` rather than `:Octo octo actions`

pwntester:
Re: actions with arguments, I prefer the second option, should be already the case for most actions

ben-davis:
@pwntester Yeah I went back and forth on `Octo actions` vs `Octo octo actions`. But now that I write that out it's obvious how much cleaner `Octo actions` is. Just updated the PR.

pwntester (approved):

pwntester:
Thanks!

tami5:
😍 Thank you

PR #251, opened 2022-01-24
let browser cmd open current PR if not specified

pgherveou:

Not sure if this is the best way to achieve this, but this would be a useful feature for me 
I can just open vim and type `:Octo pr browser` and it takes me to the current PR associated to my branch

NONE

### Describe how you did it

👇 

### Describe how to verify it

use `:Octo pr browser` after opening vim to navigate to PR

### Special notes for reviews

Since this is async it would be nice to print some kind of loading message first while we are waiting for gh to return

### DIFF
--- a/lua/octo/navigation.lua
+++ b/lua/octo/navigation.lua
@@ -11,6 +11,11 @@ function M.open_in_browser(kind, repo, number)
     local bufnr = vim.api.nvim_get_current_buf()
     local buffer = octo_buffers[bufnr]
     if not buffer then
+      utils.get_pull_request_for_current_branch(function(pr)
+        if pr ~= nil then
+          M.open_in_browser("pr", pr.repo, pr.number)
+        end
+      end)
       return
     end
     if buffer:isPullRequest() then


pwntester:
Thanks, I like the idea but I find `Octo pr browser` confusing since the rest of `Octo pr` actions are for PR buffers. Perhaps `Octo current_pr browser` or `Octo cwd_pr browser`. And we can also add `Octo cwd_pr open` to open it in octo. Im open to other ideas

pgherveou:
sorry for the slow resp here. Agree with you it does not fit well with the other commands.
a current_pr or load_current maybe as you suggested would be nice
in the meantime added myself a small bash function to load the current PR in the browser, if this can be useful to anyone...
```bash
browsePR(){
  BRANCH=$(git branch --show-current)
  PR=$(gh pr list --json "headRefName,url" --jq . | jq -r ".[] | select(.headRefName == \"$BRANCH\") | .url")
  open $PR
} 
```

disrupted:
@pgherveou perhaps I missed something, but doesn't `gh pr view --web` do what you want? https://cli.github.com/manual/gh_pr_view

pgherveou:
You need to first select a PR right? I was trying to quickly access the current pr from a single command

disrupted:
you just need to checkout your PR branch first

pgherveou:
ah nice, sorry missed your point earlier, this is much simpler indeed,
thanks for your comment

PR #271, opened 2022-03-29
Allow users to change the submit_win mappings

### LINKED ISSUE #270: Allow user to use custom submit_win mappings
## Issue Description

Type: *bug report*

I have my Tmux configured to use `<C-a>` for the prefix key which conflicts with the approve action in the review submission window. When I attempted to use a different mapping it didn't appear to change. 
I'm assuming this is a bug (not intentional) as it seems like the user's keymaps are not used for submit_win and review_diff here :
https://github.com/pwntester/octo.nvim/blob/5e461b944fbf9b6207cf06102ca09fd7778854f7/lua/octo/config.lua#L163-L168

I expected to be able to use my own submit_win mappings.

Also, thanks for the plugin 🙏 . It's very useful!

jarviliam:
To allow users to provide their own submit_win keymaps

Fixes #270 

### Describe how you did it
Added the missing user mapping override

### Describe how to verify it
Provide a custom submit_win mapping to the configuration and expect it to be used in the submit window


pwntester:
Thanks!

PR #278, opened 2022-04-06
feat: provide entry filename for better integration with nvim

muniter:

By providing the filename as part of the entry, now a user can load multiple
entries from Telescope to the quickfix/loclist and actually move through them
and making Octo fill the buffers by triggering the BufReadCmd.

I don't see one, but I think this is an actual bug. Before this code if you tried filling the quickfix list you wouldn't get anything useful.

### Describe how you did it

Break up editing function into uri generation and editing the uri.

### Describe how to verify it

1. `Octo issue list`
2. Select multiple with telescope
3. Use send to quickfix/loclist action.
4. Now navigation of the quickfix list has the buffer names, and Octo fills them.

### Special notes for reviews

I tested pull requests and issues, but not the repo one. Can't seem to make octo show me the repositores I own. Or perhaps I made a mistake?

### DIFF
--- a/lua/octo/pickers/telescope/entry_maker.lua
+++ b/lua/octo/pickers/telescope/entry_maker.lua
@@ -48,8 +48,16 @@ function M.gen_from_issue(max_number, print_repo)
     if not obj or vim.tbl_isempty(obj) then
       return nil
     end
+    local kind = obj.__typename == "Issue" and "issue" or "pull_request"
+    local filename
+    if kind == "issue" then
+      filename = utils.get_issue_uri(obj.repository.nameWithOwner, obj.number)
+    else
+      filename = utils.get_pull_request_uri(obj.repository.nameWithOwner, obj.number)
+    end
     return {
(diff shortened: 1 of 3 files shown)


pwntester (approved):

pwntester:
Thanks! Havent used the Telescope to QF action yet, but this sounds pretty cool. Can you please take a look to the stylua check?

jlopezgr:
The linter was complaining about files I didn't touch, but I wen ahead and added those changes to the commit. Is that ok?

muniter:
The linter was complaining about files I didn't touch, but I wen ahead and added those changes to the commit. Is that ok?

pwntester (approved):

pwntester:
> The linter was complaining about files I didn't touch, but I wen ahead and added those changes to the commit. Is that ok?

Yep, those are checks that keep failing, need to take a look and fix them

Thanks again for the contribution, keep them comming!

Issue #282, opened 2022-04-11
Migration to Teal with GitLab Support


l0nax:
## Issue Description

Type: "feature request"

### Describe what happened (or what feature you want)

Since I really love this plugin but mostly work with GitLab, I decided to start the migration to Teal allowing us to easily implement gitlab (or any other Host provider).

See #138 

Since the code base is not small anymore, I did not migrate the whole project to _Teal_, instead, only the parts which are relevant for the implementation are migrated/ refactored.

Currently, only GitLab issues are implemented.

pwntester:
Hi @l0nax thanks for the PR. Can you elaborate on why gitlab provider cannot be supported in plain lua? Im ok adding support for GitLab or other providers but Im not considering migrating to teal

brendalf:
Hello there. I agree with @pwntester, the gitlab support should use the glab tool as backend, which is a non official gitlab cli, akin to gh cli.
In fairness, I started looking at this because I use gitlab in my company and Octo is awesome.
I'd love to work in that integration, WDYT @pwntester?
Can I open a MR with some raw work to kick off some discussions?

pwntester:
Sure, that sounds great!

sunzoje:
Glab is now officially adopted by GitLab. https://github.com/profclems/glab/issues/983

brendalf:
Hi there. Sorry for taking so long. I will open the MR this weekend.

l0nax:
@sunzoje I already converted roughly about 70% of the code base to lua. I will push the changes this weekend.

Daemoen:
Any update on this ?

Issue #138, opened 2021-04-11
Feature Request: Gitlab Support


freexploit:
Is there any plans to support gitlab on any level?

weilbith:
I think that's not easily possible. This plugin builds fully  around the GitHub CLI tool. I don't think they will aim to support GitLab as well. On the other side the architecture of this plugin is not structured so that it would be easily possible to add an alternative "endpoint". Especially because the data objects of the GitHub API are spread all other the place.

@pwntester your statement? :thinking:

pwntester:
The code is tightly coupled to the objects returned by the GitHub GraphQL API. It should be possible to refactor the code so that the gh module invokes different queries for GitLab and then maps the returning objects to the ones expected by Octo, but that's something Im not planning to do in the short term. PRs are welcome though :)

AckslD:
@weilbith @pwntester gitlab support would be great! To my knowledge the [glab](https://github.com/profclems/glab)-cli has a quite similar API to `gh`. I don't know anything about the internals of `octo` but maybe a similar approach to how it's done in [cmp-git](https://github.com/petertriho/cmp-git) could be taken? Has anyone look into this further since above? :)

weilbith:
I just checked out the GraphQL reference of both APIs. On a first gaze they look quite similar. But some details are different. Here and there is a field missing or has slightly different type. So I think the devil is in the details.
Unfortunately there are no tests to just try it and exchange the executed binary. You would need to test everything manually and really have an eye for the details.
Though I agree with @pwntester that this plugin should actually be refactored in such a case and put an abstract interface in place. This is necessary to decouple the core functionality of this plugin from the dependencies that are used to communicate with the repository service provider. This interface would then be implemented for GitHub and GitLab. In theory also others like Mango. An important thing here is that we would need to add custom struct definitions for internal. Atm the foreign objects get used all other the place and create an huge dependency issue. But LuaJIT is also untyped and it is little nightmare to use proper typing here. At least in a developer supportive manner. This would require to use Teal or alternatives.
After all: it might work semi optimal with minimal changes by just using a different binary. The proper solution requires an immense amount of refactoring. I would love to see this. But I think all parties lack in time to do so.

PR #283, opened 2022-04-11
Gitlab with teal

l0nax:

* #138 (Partially)
* Ref: #282 

### Special notes for reviews

I've removed/ moved all the content from `lua/` to `src/` and _did not_ commit the new content of `lua/` otherwise it would be difficult to review what has changed (in the lua code).

If merged onto `master` the code has to be "compiled" and put into the `lua/` directory.

The code has to be "compiled" back to lua since we do not want to break the caching mechanism (slowing down the startup of Nvim):
```bash
~$ rm -rf lua/*
~$ tl build
~$ rsync -zarv --include="*/" --include="*.lua" --exclude="*" "src/" "lua/"
```


pwntester:
Im sorry, but any GitLab support should be added in pure lua since Im not considering moving the project to Teal at this point

farbodsz:
That's fair. @l0nax , do you think you'll work on a Lua-only version for this feature? If not, I'm happy to give it a go

brendalf:
Hi folks, I'm almost there. I will open the MR this weekend.

globin:
Any news on this?

PR #289, opened 2022-05-11
Fix broken remote detection

choplin:

The current implementation of `utils.get_remote_name`, introduced via https://github.com/pwntester/octo.nvim/commit/cd8bdcbfca4dd57b38fe25ba570a2e3f51b1f77b, returns a wrong value containing a domain name, e.g. `github.com/pwntester/octo.nvim` ,  which subsequently makes many commands not able to function properly, including `Octo issue list` , `Octo pr list` etc. 

None

### Describe how you did it

Strip domain name before returning the value.

e.g. "github.com/pwntester/octo.nvim" to "pwntester/octo.nvim"

### Special notes for reviews

I'm not sure what kind of URL is expected in the  ":/" or ":" separator case. Please tell me if this PR does not work properly for those cases.

### DIFF
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -166,7 +166,9 @@ function M.get_remote_name()
       else
         return
       end
-      return string.gsub(vim.split(url, sep)[2], ".git$", "")
+      return table.concat({
+        unpack(vim.split(string.gsub(vim.split(url, sep)[2], ".git$", ""), "/"), 2)
+      }, "/")
     end
   end
 end


pwntester:
Thanks, for sending a fix. I implemented it a little diffefently in https://github.com/pwntester/octo.nvim/commit/7fbce041ac2b9f1c97208a3653cc1af3522a5acf can you please check if it works for you?

choplin:
Thank you for letting me know it. But, after pulling the commit, `get_remote_name` returns null unexpectedly. This is probably because the remote url I tested is like `ssh://git@github.com/pwntester/octo.nvim.git`, which does not match any defined condition. While it valid ssh URL for git and it is able to access git hub repository via this kind of URL, [GitHub document](https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories) says that they support ssh URL only in the form of `git@github.com:user/repo.git`, so I'll leave it to you whether support `ssh://` style URL.

pwntester:
Got it, I added support for ssh:// remote urls too, thanks for reporting and for the PR!

Issue #276, opened 2022-04-01
Authentication Required even though already logged in


protex:
## Issue Description

Type: Bug Report

### Describe what happened (or what feature you want)
I'm attempting to use this with an enterprise account. I've set the hostname correctly in the config and also authenticated via https (I've also tried SSH). However, when I run `Oct pr list` (or any other command that requires authentication) I get this message:
```
gh: This endpoint requires you to be authenticated. (HTTP 401)
```

When I run `gh auth status`, I get this output showing me I'm logged in:
```
<enterprise-url>
  ✓ Logged in to <enterprise-url> as <username> (/Users/<username>/.config/gh/hosts.yml)
  ✓ Git operations for <enterprise-url> configured to use https protocol.
  ✓ Token: *******************
```
### Describe what you expected to happen
I except to be able to list pr's and do other things that require authentication when I am logged in

### How to reproduce it (as minimally and precisely as possible)

1. log in using `gh auth login --hostname <enterprise-url>`
2. Open neovim within git repo
3. run `Octo pr list`

### Tell us your environment
OSX 11.6.4 (intel)

pwntester:
Can you try re-authenticating using `GITHUB_TOKEN= gh auth login`

protex:
No luck, still get the same error. I also realized I can run `:!gh pr list` and that also works. It's just when I run through the octo command that it doesn't work.

bzub:
I have this issue also. It seems as though the problem only happens if there are multiple hosts in my `~/.config/gh/hosts.yaml`. A workaround is to comment out my `github.com` host if I'm using Octo with my enterprise git repo.

PR #292, opened 2022-05-17
fix: specify hostname if remote is not github.com

bzub:
`gh api` gets confused if auth for a github.com host is configured, but
the repository remote uses an enterprise github host. We avoid this by
finding the host from the remotes and specify it via flag.

Fixes: https://github.com/pwntester/octo.nvim/issues/276

### DIFF
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -182,7 +182,7 @@ function M.get_remote()
         goto continue
       end
       return {
-        host = host,
+        host = vim.split(host, "@")[#vim.split(host, "@")],
         repo = string.gsub(repo, ".git$", ""),
       }
     end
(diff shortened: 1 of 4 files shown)


lua/octo/utils.lua
```diff
@@ -155,26 +159,45 @@ function M.get_remote_name()
...
+        host = vim.split(parts[1], "@")[2]
+        local repo_chunks = vim.split(parts[2], "/")
+        repo = repo_chunks[1] .. "/" .. repo_chunks[2]
       else
-        return
+        goto continue
       end
-      return string.gsub(repo, ".git$", "")
+      return {
+        host = host,
```
pwntester, line 185:
```suggestion
        host = vim.split(host, "@")[#vim.split(host, "@")],
```

We need to remove the username from the host chunk

pwntester (changes requested):

pwntester:
Thanks for making Octo better!

lua/octo/utils.lua
bzub: Fixed.

PR #297, opened 2022-05-27
allow web devicons to be an optional dependency

quentin-fox:

In the README, it looks like the `file_picker.use_icons` field could be set to false to make the web devicons an optional dependency, but the line that reads from this config property was commented out.

NONE

### Describe how you did it

Just uncommented the relevant line out.

### Describe how to verify it

Uninstall the nvim-web-devicons package from the nvim runtime, and open the file picker. It should still work without a lua `require` stack trace appearing in the messages pane, provided that the `file_picker.use_icons` config property is set to `false`.


pwntester:
Thanks for the PR!

PR #322, opened 2022-08-24
Update vim.fn.trim for nightly neovim

aaronhallaert:

I updated my neovim to the `nightly` branch and broke `octo.nvim` when opening a PR.
Apparently `vim.fn.trim` is renamed to `vim.trim`.

### Describe how you did it

Replace all occurences of `vim.fn.trim` to `vim.trim`

### Describe how to verify it

Open a PR with octo.nvim

### Special notes for reviews

I'm not sure if this PR should be merged for now since this is a fix for a nightly version of Neovim.

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -664,7 +664,7 @@ function M.create_issue(repo)
   )
   if choice == 1 then
     local lines = vim.api.nvim_buf_get_lines(0, 0, -1, true)
-    body = utils.escape_char(vim.trim(table.concat(lines, "\n")))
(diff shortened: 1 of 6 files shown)


pwntester:
Thanks for the PR, and sorry for the late response (still on vacations). Im sorry I cannot merge the PR as is since it would break older versions. Since I still dont have access to a laptop, could you make a `trim` function in the `utils` module that calls either `vim.fn.trim` or `vim.trim` depending on which one exists?

pwntester:
Thank you @aaronhallaert !

PR #346, opened 2022-12-09
add ability to set computed env vars for gh cli

### LINKED ISSUE #345: [FEATURE REQUEST]: Get `GITHUB_TOKEN` from a function
I see that in https://github.com/pwntester/octo.nvim/blob/master/lua/octo/gh/init.lua 

you grab `GITHUB_TOKEN` from `vim.env`. I would like to use the 1Password CLI to populate the `GITHUB_TOKEN`.

Currently I have it set up like so (using the `op.nvim` 1Password Neovim plugin):

```lua
local github_token = require('op.api').item.get({ '[item uuid here]', '--fields', 'token' })[1]
if not github_token or not vim.startswith(github_token, 'ghp_') then
  error('Failed to get GitHub token.')
end
vim.env.GITHUB_TOKEN = github_token
require('octo').setup()
```

However, this means I am prompted for auth immediately, instead of only when `octo.nvim` actually goes to run the `gh` CLI.

Would it be possible to have an option in `setup` like so:

```lua
require('octo').setup({
  get_gh_token = function()
    local github_token = require('op.api').item.get({ '[item uuid here]', '--fields', 'token' })[1]
    if not github_token or not vim.startswith(github_token, 'ghp_') then
      error('Failed to get GitHub token.')
    end
    return github_token
  end,
})
```

mrjones2014:

Adds a hook to get computed environment variables when running the GitHub CLI. It is needed, for example, to allow you to set `GITHUB_TOKEN` from the 1Password CLI.

Fixes #345 

### Describe how you did it

Added a config option that allows user to pass a function that returns a table to merge with `env_vars`

### Describe how to verify it

```lua
require('octo').setup({
  get_env = function()
    get_env = function() return { GITHUB_TOKEN = 'your token here' } end,
  end,
})
```

This config should authenticate the GitHub CLI via octo.nvim

### Special notes for reviews

Here's how I've configured mine, integrating with the `op.nvim` plugin, using its API:

```lua
require('octo').setup({
  get_env = function()
    local github_token = require('op.api').item.get({ '[item uuid here]', '--fields', 'token' })[1]
    if not github_token or not vim.startswith(github_token, 'ghp_') then
      error('Failed to get GitHub token.')
    end

    return { GITHUB_TOKEN = github_token }
  end,
})
```


pwntester:
Thanks for the PR!

Although in your specific use case it makes sense for `get_env` to be a function, I would expect most users would suffice to use a table. Can you make the following changes?

- rename `get_env` to `gh_env`
- handle both function and table cases
- document the new `gh_env` config option

mrjones2014:
Done!

pwntester (approved):

pwntester:
Awesome, thanks!

PR #348, opened 2022-12-12
Fix the floating window size calculation

yen3:
If the number of max lines >= 1, it calculates a float number. It could apply to `vim.api.nvim_open_win`. Add the converting action to make sure the final result type is int.

Pass the size with correct type to create floating window.

NONE

### Describe how you did it

The result of the value may not be an integer, add the convert to make sure it's integer type.

### Describe how to verify it

I verify in my local environment with `:Octo pr checks`


pwntester (approved):

pwntester:
Thanks!

PR #370, opened 2023-02-05
feature: add glab support

### LINKED ISSUE #282: Migration to Teal with GitLab Support

brendalf:
This MR adds support to Gitlab CLI.

Fixes #282

### Describe how you did it
This MR implements the support to Gitlab CI by moving the current calls to `gh.run` to a backend interface.

The idea is that in the future we can add support to which cli we want by just adding a new `backend` that implements 
the same functions that `gh` currently does.

By default, if the `backend` doesn't implement some action, a notification is triggered to warn the user.

### Describe how to verify it
Still WiP

### Special notes for reviews
This MR is still WiP.

I'm opening it so we can start discussing the implementation plan.

Do you agree with the decision of adding the `backend` interface?

Currently action plan:
- [x] Add backend interface.
- [ ] Migrate remaining `gh` and `graphql` usage to backend functions inside `backend.gh`.
- [ ] Write similar `backend.gh` functions to `backend.glab`.
- [ ] Write a function to determine which backend to use.


brendalf:
Hi @pwntester. WDYT about this plan?

pwntester:
Hi @brendalf Thanks a lot for the PR. The plan sounds just right, but my guess is that `glab` GraphQL endpoints are going to return object trees that are completely different from the `gh` ones. That means that we are going to need a **conversion** function to make `glab` responses processable by Octo. Not sure how difficult that is going to be or if its going to be even possible

brendalf:
Great. So I will try implementing the glab respective methods before migrating gh calls to the new format. This way I can address your concern first.

brendalf:
At lot of things changed in my life and I was unable to proceed with the work, like becoming a father and moving to the Netherlands. Now that we have #530, I will close this one, but I happy to contribute to other features after one is merged (if that happens).

Issue #369, opened 2023-02-02
How do I configure the timeout ? 


shivanthzen:
## Issue Description

Some times I have slow network and working on a large repository gives me timeout 5s when pushing a branch/ creating a PR.  Is there a way I can configure the timeout in octo ?

Issue #375, opened 2023-02-27
Is there a way to increase the timeout for a git push command ? like in configurations...


Sanix-Darker:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)

When am trying to create a new Pull Request, and because I have to provide a key-pass-phrase, am getting this error while trying to push : 
```
Pushing 'xxxx' to 'origin:xxxx' ...
Error executing Lua callback: ...dk/.config/nvim/plugged/plenary.nvim/lua/plenary/job.lua:499: 'git push origin xxxx:xxxx' was unable to complete in 5000 ms
stack traceback:
        [C]: in function 'error'
        ...dk/.config/nvim/plugged/plenary.nvim/lua/plenary/job.lua:499: in function 'wait'
        ...dk/.config/nvim/plugged/plenary.nvim/lua/plenary/job.lua:453: in function 'sync'
        .../dk/.config/nvim/plugged/octo.nvim/lua/octo/commands.lua:806: in function 'create_pr'
        .../dk/.config/nvim/plugged/octo.nvim/lua/octo/commands.lua:93: in function 'a'
        .../dk/.config/nvim/plugged/octo.nvim/lua/octo/commands.lua:356: in function 'octo'
        .../dk/.config/nvim/plugged/octo.nvim/lua/octo/commands.lua:16: in function <.../dk/.config/nvim/plugged/octo.nvim/lua/octo/commands.lua:15>
```
I guess, it's because the timeout is too short ?

### Describe what you expected to happen

Maybe a big timeout ? Or a way to increase that timing in configuration ? 🤔 

### How to reproduce it (as minimally and precisely as possible)

1. SetUp a git profile with ssh and keys protected with a key pass phrase
2. inside a new repo, try to create a new Pull Request
3. 

### Tell us your environment
Am on ubuntu, with nvim nightly : 
```
NVIM v0.9.0-dev-974+g631775c05
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #376, opened 2023-03-01
feat: added a timeout param to be able to configure long process time

Sanix-Darker:

Add `timeout` param to config to have a custom timeout for processes that can take time.

Resolves this https://github.com/pwntester/octo.nvim/issues/375 and should probably resolves : https://github.com/pwntester/octo.nvim/issues/369

### Describe how you did it

I just added a new entry from the configuration, and call it inside the FileEntry function


pwntester:
Thanks for the contribution!

PR #389, opened 2023-04-26
Use hostname flag to get user name

### LINKED ISSUE #388: Bug: Username for Comments do not match for local repo config
## Issue Description

bug report: Username for comments do not match local repo config

While writing a comment i recognized that it uses the username i have on github.com instead the one i have for the GHE instance on which the local repo is hosted.

As far as i can tell the username is derived from here:

https://github.com/pwntester/octo.nvim/blob/master/lua/octo/gh/init.lua#L45 which takes only the first occurrence in consideration. 

The username for comments etc should be local to the repo, eg if a github.com repo a github.com username, on GHE repo my GHE username.

1. have 2 host configured in ~/.config/gh/hosts.yml
2. write a comment in GHE repo, username will be the one from github.com

### Tell us your environment
Ubuntu 22.04 neovim nightly

in https://github.com/pwntester/octo.nvim/blob/master/lua/octo/gh/init.lua#L45 `gh auth status` is used to derive the username.

Even if github.com is second in ~/.config/gh/hosts.yml it will appear first in the list, from which then the username is derived

`gh auth status` also takes a `--hostname` flag which could be used to get the username for the local repo.

Shinzu:

With this pr the local repo hostname is used to determine the Username for eg comments and other operations that need this username.

Fixes #388 

### Describe how you did it

To determine the Username `gh auth status` is used, which also takes the flag `--hostname`. I added this flag to the `gh auth status` command with the remote host i get from the `get_remote_host` function from `octo.utils`.

If you have 2 hosts configured in ~/.config/gh/hosts.yml like github.com and a GHE instance, the current implementation only takes the first  output in consideration to determine the username, which in my test is always github.com even if this is the second host configured in ~/.config/gh/hosts.yml.

With adding the `--hostname` flag we only consider the local repos remote for determine the username.

### Describe how to verify it

You have 2 host configured in ~/.config/gh/hosts.yml and while create a comment on an issue the username on the host the repo is hosted on (github.com or GHE instance) is used in the comment.

### DIFF
--- a/lua/octo/gh/init.lua
+++ b/lua/octo/gh/init.lua
@@ -42,11 +42,11 @@ local function get_env()
 end
 
 -- uses GH to get the name of the authenticated user
-function M.get_user_name()
+function M.get_user_name(remote_hostname)
   local job = Job:new {
     enable_recording = true,
     command = "gh",
-    args = { "auth", "status" },
+    args = { "auth", "status" , "--hostname", remote_hostname },
     env = get_env(),
   }
   job:sync()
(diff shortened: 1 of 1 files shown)


pwntester:
LGTM, thanks for the contribution!

PR #399, opened 2023-05-26
refactor: call git only once to get remotes

delphinus:

To get remotes, the current build calls `git remote show` and `git remote <name>` multi times. Now it calls `git remote -v` only once and can get all results from this.

### Special notes for reviews


pwntester:
Thanks!

PR #407, opened 2023-06-08
add mappings for selecting first and last entries

tohalla:
First of, thanks for this awesome plugin. It helps with the review process immensely!
 
Especially with larger pull requests with many file changes, I often find a need to return beginning or end of the file list. I know that the issue can be somewhat resolved via navigating to the file list and selecting the file from there, but I'm sure having intuitive mappings around for navigation would feel more natural (at least for me).  

NONE

### Describe how to verify it

Use `]Q` `[Q` for navigating.


pwntester (approved):

pwntester:
Thanks for the contribution!

PR #409, opened 2023-06-10
feature: Show builtin actions when no action is provided

### LINKED ISSUE #377: Octo builtin
As a user, when I run `Octo` I want to see a list of built-in commands available to run.

It would be great if we could borrow from what `Telescope` does and show a Telescope popup of all available built-in commands from Octo if it's run on it's own. It currently just says "Missing Argument"

Screenshot 2023-03-01 at 12 07 39 PM

jarviliam:
Provides a QOL feature to show the list of builtin actions when no action is provided.

Fixes #377 

### Describe how you did it
Added the config option `enable_builtin` which when enabled runs `Octo actions` when the user runs the default `Octo` command.

### Describe how to verify it
When turned off the behavior of `Octo` should be the same. When enabled it should run `Octo actions`

### Special notes for reviews
I've set the default behavior of it to be off but maybe it should just be on by default and just remove the error message when no action is provided? 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ use {
 ```lua
 require"octo".setup({
   use_local_fs = false,                    -- use local files on right side of reviews
+  enable_builtin = false,                  -- shows a list of builtin actions when no action is provided
   default_remote = {"upstream", "origin"}; -- order to try remotes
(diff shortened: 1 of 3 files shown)


pwntester (approved):

pwntester:
Thanks!

PR #422, opened 2023-09-27
fix: error opening current issue/pr/repo in browser when using Github Enterprise

f3nrir92:

When plugin is configured to use Github Enterprise instance there is an issue related to opening PR/repo/issue from current buffer. It's unnecessary to use a `repo` argument to `gh` command in this case, so we can skip it.
NONE

### Describe how you did it
I just removed `-R` argument from the call of `gh` command

### DIFF
--- a/lua/octo/navigation.lua
+++ b/lua/octo/navigation.lua
@@ -24,6 +24,11 @@ end
 
 function M.open_in_browser(kind, repo, number)
   local cmd
+  local remote = utils.get_remote_host()
+  if not remote then
+     utils.error "Cannot find repo remote host"
+     return
+  end
   if not kind and not repo then
     local bufnr = vim.api.nvim_get_current_buf()
     local buffer = octo_buffers[bufnr]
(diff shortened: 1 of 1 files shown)


lua/octo/navigation.lua
```diff
@@ -31,11 +31,11 @@ function M.open_in_browser(kind, repo, number)
       return
     end
     if buffer:isPullRequest() then
-      cmd = string.format("gh pr view --web -R %s %d", buffer.repo, buffer.number)
+      cmd = string.format("gh pr view --web %d", buffer.number)
```
pwntester, line 34: Will this work for GitHub.com though? I guess it will if your CWD is in the repo you want to browse, but will probably fail otherwise

pwntester (changes requested):

lua/octo/navigation.lua
f3nrir92: Hm... I'm never tried to use it this way, will return later with changes. Thanks for notice

f3nrir92:
@pwntester thanks for suggestion! 
It was fixed and squashed :)

pwntester (approved):

pwntester:
Awesome, thanks a lot!

PR #425, opened 2023-10-05
Improve background color

joshmedeski:
Closes #424 

Adjusting the `get_background_color_of_highlight_group` function to support cleared highlight groups.

- Falling back to `highlight_group_normal.foreground` is both confusing and makes it impossible to overwrite the default background value to a null value. So I've removed it.
- Removed the fallback else if background color isn't defined. In my case, I don't want there to be a background color so I'm clearing the `OctoEditable` highlight but the previous logic would still add a color back.

Resolves https://github.com/pwntester/octo.nvim/discussions/424

### Special notes for reviews

Please let me know if the black background was intentional. And if so, what I would need to do to allow the `get_background_color_of_highlight_group` to return a cleared value so I can get a transparent UI.

### DIFF
--- a/lua/octo/ui/colors.lua
+++ b/lua/octo/ui/colors.lua
@@ -177,13 +177,9 @@ end
 function M.get_background_color_of_highlight_group(highlight_group_name)
   local highlight_group = vim.api.nvim_get_hl_by_name(highlight_group_name, true)
   local highlight_group_normal = vim.api.nvim_get_hl_by_name("Normal", true)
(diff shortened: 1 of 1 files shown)


pwntester:
Thanks! it looks good to me. I guess I was using black as the fall safe option, but it makes sense to allow the color to be cleared

joshmedeski:
No, thank YOU! I love using this plugin, such a time saver.

And thanks for the quick merge.

PR #435, opened 2023-10-16
Fix wrong argument in load buffer function

kezhenxu94:

 the function `M.load_buffer` accepts the buffer number but the reload function passes in several arguments, causing the `Octo pr reload` always to fail 

https://github.com/pwntester/octo.nvim/blob/bd53dee3ab1bf13882e3e5013f355d802b0a76ad/lua/octo/init.lua#L64


pwntester:
Good catch! thanks for the contribution!

PR #436, opened 2023-10-17
feat: projects v2 support

milogert:

This PR implements the projects v2 interface as well as the picker for fzf-lua.

Adds:

- Projects v2 support as well as the fzf-lua picker addition.
  - The interface for the new command changed a bit. Instead of `add`, `move`, and `remove` I went with `set` and `remove`. I The flow of projects v2 changed a bit and `add` and `move` felt a little too similar.
- A config option `default_to_projects_v2`, which makes the `card` command default to use projects v2. Both versions are still available under the `cardlegacy` and `cardv2` commands, regardless of this setting.

Addresses https://github.com/pwntester/octo.nvim/discussions/414

### Describe how you did it

Followed the pattern of the other cards command. As noted above I changed the interface for the new command slightly, but to something that made sense to me and how projects v2 flowed.

### Describe how to verify it

Mess around with projects v2 inside Octo. Everything should work as expected.

### Special notes for reviews

I added two small additions as riders:

- Adds query timeout to searching in fzf-lua to make search more responsive.
- Adds draft PR special highlighting when viewing an pr list in fzf-lua.

These are tiny changes that are just QoL adds for fzf-lua.

Also I updated the README to account for some fzf-lua changes as well.


pwntester:
Thanks for the PR, looks great! Will review ASAP

Cih2001:
Cant wait to see this merged guys, the only reason I have Telescope installed is Octo. Thanks for this.

milogert:
@Cih2001 fwiw, fzf-lua is already supported in Octo.nvim. This PR _extends_ that support (even beyond Telescope's support) with Projects V2 additions. The bulk of Octo works just fine though

pwntester:
@milogert sorry for the late review. Can you please rebase against master?

milogert:
I believe I am.. Right? I might be missing something though. My GitHub ui says that I'm just  8 commits ahead of master and no commits behind it

pwntester:
Oh, you are, sorry. It looks good to me, but can you please add placeholders for the missing pickers for [telescope](https://github.com/pwntester/octo.nvim/blob/master/lua/octo/pickers/telescope/provider.lua#L1059) and [vim-clap](https://github.com/pwntester/octo.nvim/blob/master/lua/octo/pickers/vim-clap/provider.lua#L26)?

milogert:
Yep, just pushed changes up over lunch :) Thanks for the eyes!

pwntester:
Thanks you for the contribution!

vieteh:
Hello,
I see this PR has already been reverted by https://github.com/pwntester/octo.nvim/pull/474
I guess because of the error about `gh` token not having the `projects` permission.

Just wanna share that I can overcome that.
I manually created a token with the permission. Then `gh auth login`, and use the option to fill an existing token instead of auto generating.

So I think this feature should be updated to be auto-available when the token has the permission.

tiev:
Hello,
I see this PR has already been reverted by https://github.com/pwntester/octo.nvim/pull/474
I guess because of the error about `gh` token not having the `projects` permission.

Just wanna share that I can overcome that.
I manually created a token with the permission. Then `gh auth login`, and use the option to fill an existing token instead of auto generating.

So I think this feature should be updated to be auto-available when the token has the permission.

pwntester:
Thanks @tiev a couple of alternatives are being discussed [here](https://github.com/pwntester/octo.nvim/issues/473#issuecomment-1888632697), feel free to chime in, but the idea of reverting this PR is to avoid people not using projects v2 to have to upgrade their tokens.

Issue #412, opened 2023-06-17
highlighting of badges is funky?


mizlan:
## Issue Description

Type: *bug report* or *feature request*

### Describe what happened (or what feature you want)

<img width="438" alt="Screen Shot 2023-06-17 at 9 02 14 AM" src="https://github.com/pwntester/octo.nvim/assets/44309097/a68fc3f0-6dbb-4c7d-a7b4-c0398791ef78">

The badges use the special char but it's highlighted wrong.

### Describe what you expected to happen

The badge should be properly highlighted (it seems Octo should tack on a default highlight group for this)

### How to reproduce it (as minimally and precisely as possible)

1. Use Octo.nvim
2. Set colorscheme to Rose-pine or Everforest , seoul256, tokyonight (tested with those 4)
3. see it

### Tell us your environment

unnecessary

pwntester:
Hi, probably those themes are missing the Octo highlight groups. What are your colours for:
- OctoBubbleBlue
- OctoBubbleDelimiterBlue
- OctoBubbleYellow
- OctoBubbleDelimiterYellow
- OctoBubbleGreen
- OctoBubbleDelimiterGreen
- OctoBubbleRed
- OctoBubbleDelimiterRed

rodhash:
Similar issue happening here, in my case it seems to be a bit more broken.

<img width="452" alt="image" src="https://github.com/pwntester/octo.nvim/assets/29671981/726ca678-2c4d-4d10-bc9b-35c7840f7578">



iTerm `3.4.22`
Nvim `v0.9.4`
Octo `604fad12`


These are my colors (my custom teme):

<img width="697" alt="image" src="https://github.com/pwntester/octo.nvim/assets/29671981/67460398-a903-460d-8ac7-572395bf2ad8">


I also tried other colorschemes such as gruvbox and tried using kitty instead but same issue. I even installed 4 or 5 different NF fonts but no luck so far.

any idea?

rodhash:
Here a print screen w/ Kitty, Gruvbox, latest neovim nightly and a new NF:

<img width="687" alt="image" src="https://github.com/pwntester/octo.nvim/assets/29671981/31cef0ce-7777-49bc-930b-f3d3d195ee54">

kezhenxu94:
For everyone in this thread, I've filed a PR (https://github.com/pwntester/octo.nvim/pull/456) to fix this, in case you want to verify.

rodhash:
[...]

PR #456, opened 2023-11-24
Fix wrong bubble background colors

kezhenxu94:

This PR fixes the wrong bubble background colors, which should be introduced in https://github.com/pwntester/octo.nvim/pull/433

Possibly fixes https://github.com/pwntester/octo.nvim/issues/412

### Describe how you did it

Looks like `vim.api.nvim_get_hl` returns a different structure that contains `{ bg, fg }` instead of `{ background, foreground }`, this PR fixes the property names.

### Describe how to verify it

- Before this PR:

image

- After this PR:

image

### DIFF
--- a/lua/octo/ui/colors.lua
+++ b/lua/octo/ui/colors.lua
@@ -175,9 +175,9 @@ local function color_is_bright(r, g, b)
 end
 
 function M.get_background_color_of_highlight_group(highlight_group_name)
-  local highlight_group = vim.api.nvim_get_hl(0, { name = highlight_group_name })
-  local highlight_group_normal = vim.api.nvim_get_hl(0, { name = "Normal" })
-  local background_color = highlight_group.background or highlight_group_normal.background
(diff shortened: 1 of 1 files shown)


rodhash:
Hi

It seems 'CreatedBy' is having similar issues:

image

rodhash:
Well in my test repo the "Created user" shows up ok, no issues .. but in this example from the Neorg repo, for some reason the "Created by" is like that.. it's using the HL group OctoBubble.

pwntester:
Thanks, looks fine to me! As for the neorg issue it renders like this to me:

image

Note that user names use the `OctoUser` hl group. In my theme I use [this](https://github.com/pwntester/nautilus.nvim/blob/main/lua/nautilus/theme.lua#L474)

pwntester:
@rodhash can you confirm if your problem is related with `OctoUser` hl group please?

rodhash:
Hi @pwntester, yes this one too..

It seems to me we have two hl groups that apply to this situation, `OctoUser` as you mentioned but also `OctoBubble` .. both applies to field "Created by"

PR #464, opened 2023-12-09
Fix thread header becoming invisible after toggling resolved-state

### LINKED ISSUE #362: Thread becomes invisible after Octo thread resolve
## Issue Description

Type: *bug report* 

When resolve thread on PR overview dashboard, thread becomes invisible.

Before resolution
before_resolve

After resolution
after_resolve

Thread becomes collapsed, marks as resolved stays visible
desired

1. `Octo pr list`
2. Open PR with not resolved comments/threads
3. On PR view try to resolve thread by typing `Octo thread resolve`

### Tell us your environment
iTerm2
nvim 0.8.0
MacOs
octo plugin latest
tried with EdenEast/nightfox.nvim and projekt0n/github-nvim-theme themes.

Daniel1854:
When resolving threads via `Octo thread [un]resolve` the virtual text of the thread should be updated with the new resolved-state. Currently the text only gets cleared due to an oversight.

Fixes #362 

### Describe how you did it
Added a field, which was needed since a feature commit

### Describe how to verify it
Tested it within the timeline (`Octo pr list`)  aswell as within the review mode (`Octo review start`). Just resolved and unresolved some threads.

### Special notes for reviews
Fwiw this seems to be introduced with the commit 50704c5071e9dd4a5643f55403f146b2bbe710bc. With this commit the wrapper around writing this specific virtual_text now expects a `commit` field. 

I have added a for loop to gather the commit sha of the actual thread from the response from github. If there is an easier way, Im happy to change it. 

Sidenote: the duplicated code between resolve_thread and unresolve_thread increases with this commit. If you want, I can extract those twenty-ish lines into another function

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -561,6 +561,23 @@ function M.delete_comment()
   end
 end
 
+local function update_review_thread_header(bufnr, thread, thread_line)
+  local start_line = thread.originalStartLine ~= vim.NIL and thread.originalStartLine or thread.originalLine
+  local end_line = thread.originalLine
+  writers.write_review_thread_header(bufnr, {
+    path = thread.path,
+    start_line = start_line,
+    end_line = end_line,
+    isOutdated = thread.isOutdated,
+    isResolved = thread.isResolved,
+  }, thread_line - 2)
+  local threads = thread.pullRequest.reviewThreads.nodes
+  local review = reviews.get_current_review()
+  if review then
+    review:update_threads(threads)
+  end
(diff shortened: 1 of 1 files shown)


pwntester:
Thanks for the fix! Yep, I think it will be better to move the shared code to its own function. Can you please do that?

Daniel1854:
Sure, I extracted the logic. Let me know if there is anything else to fix

pwntester:
Nice! Thanks a lot!

Issue #466, opened 2023-12-17
Autocmd errors when opening `octo://*` buffers


scottschroeder:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)

When opening any `octo://*` path, I'd see
```
Error detected while processing BufReadCmd Autocommands for "octo://*":  
```

here is a specific example
```
:Octo pr edit 97
Error executing Lua callback: vim/_editor.lua:0: nvim_exec2()..BufReadCmd Autocommands for "octo://*": Vim(append):
stack traceback:
        [C]: in function 'nvim_exec2'
        vim/_editor.lua: in function 'cmd'
        ...ager/.local/share/nvim/lazy/octo.nvim/lua/octo/utils.lua:581: in function 'get_pull_request'
        ...r/.local/share/nvim/lazy/octo.nvim/lua/octo/commands.lua:71: in function 'a'
        ...r/.local/share/nvim/lazy/octo.nvim/lua/octo/commands.lua:377: in function 'octo'
        ...r/.local/share/nvim/lazy/octo.nvim/lua/octo/commands.lua:18: in function <...r/.local/share/nvim/lazy/octo.nvim/lua/octo/commands.lua:17>
Press ENTER or type command to continue
```
### Investigation

It looks like the match statement for extracting the username does not match output `gh`.

https://github.com/pwntester/octo.nvim/blob/4a60f50bb886572a59fde095b990fa28e2b50dba/lua/octo/gh/init.lua#L59-L60

where the output from `gh`
```
❯ gh auth status --hostname github.com
github.com
  ✓ Logged in to github.com account scottschroeder (keyring)
```

```diff
- "Logged in to [^%s]+ as ([^%s]+)"
+ "Logged in to [^%s]+ account ([^%s]+)"

```
### Tell us your environment
octo commit 4a60f50bb886572a59fde095b990fa28e2b50dba (master)

```
❯ gh --version
gh version 2.40.1 (2023-12-13)
https://github.com/cli/cli/releases/tag/v2.40.1
```

```
❯ nvim --version
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1692716794
```

### Anything else we need to know?

- This can be easily worked around by setting
```
vim.g.octo_viewer = "<your_github_user>"
```

joshmedeski:
Thanks! The work around is valid until the relate PR is merged 👍

PR #467, opened 2023-12-26
fix: Adjust get_user_name regex to match updates to the cli

jarviliam:
This commit fixes an issue with the newer version of the gh cli tool where they updated the stdout message.

Reference: https://github.com/cli/cli/commit/e806664ef736a1cba373155842112831d094b91f

Closes: #466

### DIFF
--- a/lua/octo/gh/init.lua
+++ b/lua/octo/gh/init.lua
@@ -56,8 +56,12 @@ function M.get_user_name(remote_hostname)
   job:sync()
   local stderr = table.concat(job:stderr_result(), "\n")
   local stdout = table.concat(job:result(), "\n")
+  -- Newer versions of the gh cli have a different message. See #467
   local name_err = string.match(stderr, "Logged in to [^%s]+ as ([^%s]+)")
+    or string.match(stderr, "Logged in to [^%s]+ account ([^%s]+)")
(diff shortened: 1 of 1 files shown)


pwntester (approved):

lua/octo/gh/init.lua
```diff
@@ -56,8 +56,8 @@ function M.get_user_name(remote_hostname)
   job:sync()
   local stderr = table.concat(job:stderr_result(), "\n")
   local stdout = table.concat(job:result(), "\n")
-  local name_err = string.match(stderr, "Logged in to [^%s]+ as ([^%s]+)")
-  local name_out = string.match(stdout, "Logged in to [^%s]+ as ([^%s]+)")
+  local name_err = string.match(stderr, "Logged in to [^%s]+ account ([^%s]+)")
```
pwntester, line 59: Thanks, can you make it try the old regexp if the new one fails for backwards compatibility?

pwntester (changes requested):

lua/octo/gh/init.lua
jarviliam: resolved 👍

pwntester (approved):

pwntester:
Thanks!

PR #468, opened 2023-12-27
feat: Rework configuration setup to add validation

jarviliam:
This commit reworks the config setup structure and adds Config typing, validations and tests. On config setup the user's configuration will be validated against. Any type errors raised by invalid settings will result in a user-friendly error message being shown to the user. I've lifted a majority of it from another plugin (neogit) that I've contributed to before as It seems to work well.

I've run into issues in the past where my configuration was out of date and once I updated I'd have to track down the change and apply it. With this it should be a lot more easier to see the config issue.

This also adds much needed typing annotations and configuration tests to make it easier to detect regressions/unintended changes. I still need to add validations for the mappings but this is essentially the base of it.

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -341,7 +341,7 @@ function M.octo(object, action, ...)
   end
 
   if not object then
-    if config.get_config().enable_builtin then
+    if config.values.enable_builtin then
       M.commands.actions()
     else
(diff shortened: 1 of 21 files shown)


pwntester (approved):

pwntester:
Thanks a lot @jarviliam this is an awesome contribution!

PR #470, opened 2023-12-27
feat: Use gh's resolved repo config for git remotes

jarviliam:
This commit adjusts the fetching remote function to also consider
default repositories set by the `gh` cli. When the local git repository
has `remote.origin.gh-resolved pwntester/octo.nvim` set, the resolved
repository name will be used instead of the default one.

When attempting to run any gh cli commands in a repo where there are multiple
remotes defined, it requires the user to set the default repo to use via `gh
repo set-default`. This repo is then used as the default remote for all gh
commands. This commit attempts to also utilize the same configuration.

This is needed because on my fork when I run `Octo issue list` I get the issues
from my fork and not the original repo. Now that I have set the default repo
locally, I can fetch the issues from above command.

When fetching remotes it will:
1) Fetch remotes via `git remote -v`
2) Apply user_defined ssh_aliases against the set
3) Fetch the local `gh-resolved` value
4) When returning the default remote, if a resolved value is set from step 3,
utilize that value

Let me know if theres any issues/ideas.

### DIFF
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -139,7 +139,20 @@ function M.is_blank(s)
   )
 end
 
-function M.parse_remote_url(url, aliases)
+---@class Remote
+---@field name string
+---@field host string
+---@field repo string|nil
+---@field resolved string|nil
+---@field push_url string
+---@field fetch_url string
+
+--- Parses a remote url into a Remote class
+---@param url string url of remote
+---@param name string name of remote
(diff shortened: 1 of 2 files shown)


wd60622:
Hi @jarviliam 

What is the status on this?

jarviliam:
@wd60622 Sorry for keeping it a draft. I'll close this pr out as I don't think its an issue anymore 👍

PR #477, opened 2024-01-13
Projects v2, take 2

milogert:

This PR is identical to #436 EXCEPT for the following:
 
1. During startup the scopes are cached from `gh`. I am on the fence about this particular feature, but it sort of tracks with the rest of the plugin world (i.e. restart Neovim after changes to see updates).
2. There is now a check when opening an issue or pr with Octo to see if you have the required scopes (cached during startup) for projects. If you do it inserts a fragment into the respective graphql queries. If you do not just a blank string is inserted and a printed message is shown.
3. Additionally I have added some documentation relating to this message, in the event that people who see it don't know how to add scopes off hand (like me, before I puzzled it out).
4. Finally, I added a config option to suppress the message if you don't want to see it at all.

ℹ️ The remainder of this description is just copy/pasted for documentation purposes.

This PR implements the projects v2 interface as well as the picker for fzf-lua.

Adds:

- Projects v2 support as well as the fzf-lua picker addition.
  - The interface for the new command changed a bit. Instead of `add`, `move`, and `remove` I went with `set` and `remove`. I The flow of projects v2 changed a bit and `add` and `move` felt a little too similar.
- A config option `default_to_projects_v2`, which makes the `card` command default to use projects v2. Both versions are still available under the `cardlegacy` and `cardv2` commands, regardless of this setting.

Addresses https://github.com/pwntester/octo.nvim/discussions/414

### Describe how you did it

Followed the pattern of the other cards command. As noted above I changed the interface for the new command slightly, but to something that made sense to me and how projects v2 flowed.

### Describe how to verify it

Mess around with projects v2 inside Octo. Everything should work as expected.

### Special notes for reviews

I added two small additions as riders:

- Adds query timeout to searching in fzf-lua to make search more responsive.
- Adds draft PR special highlighting when viewing an pr list in fzf-lua.

These are tiny changes that are just QoL adds for fzf-lua.

Also I updated the README to account for some fzf-lua changes as well.


pwntester:
@milogert I got some errors since the 4th property for the `pull_request_query` and `issue_query` queries was not always present. I made a global flag set on initialization and applied it on every PR/issue queries https://github.com/pwntester/octo.nvim/pull/477/commits/5f1fbf980295282b65b2e1e2f777dda7cb0fa867
Let me know if that looks good to you

lua/octo/gh/init.lua
```diff
@@ -82,7 +82,8 @@ function M.setup()
   }
   job:sync()
   local stdout = table.concat(job:result(), "\n")
-  local all_scopes = string.match(stdout, "- Token scopes: (.*)")
+  local all_scopes = string.match(stdout, " Token scopes: (.*)")
+  print(vim.inspect(all_scopes))
```
pwntester: removed in next commit

lua/octo/gh/init.lua
```diff
@@ -82,7 +82,8 @@ function M.setup()
   }
   job:sync()
   local stdout = table.concat(job:result(), "\n")
-  local all_scopes = string.match(stdout, "- Token scopes: (.*)")
+  local all_scopes = string.match(stdout, " Token scopes: (.*)")
```
pwntester, line 85: Had to remove the `-` since in my case it was a `✔`

milogert:
@pwntester that flow looks great! There was one spot that I believe you missed in my latest commit

Thanks again for taking a look at this!

pwntester:
Thanks @milogert!

PR #483, opened 2024-01-18
fix: create OctoBuffers for issue previews

### LINKED ISSUE #453: No way to identify if PR is Draft or not.
## Issue Description

In PR view you don't see if PR is in Draft state.

No way to identify if PR is Draft or not.

I expect to see line indicating that PR is Draft (if it is).

1. Run `:Octo pr list` and select PR that is draft

HampusMat:

The write_state function relies on there being a buffer in the octo_buffers table but there is none when called from the callback which is called after running Github CLI when defining a issue preview.

This causes that `[DRAFT]` isn't included in the issue preview title virtual text of draft pull requests.

A part of fixing #453.

### Describe how you did it

By making it so that before `write_state` is called, a `OctoBuffer` is created.

### Describe how to verify it

Do `:Octo pr list` in a repository with draft pull requests and you'll see the preview(s) contain `[DRAFT]`

### DIFF
--- a/lua/octo/pickers/telescope/previewers.lua
+++ b/lua/octo/pickers/telescope/previewers.lua
@@ -38,6 +38,17 @@ local issue = defaulter(function(opts)
               elseif entry.kind == "pull_request" then
                 obj = result.data.repository.pullRequest
               end
+
+              -- We have to create a new buffer since writers.write_state relies on
+              -- octo_buffers containing a buffer at a index of the current buffer number.
+              -- OctoBuffer:new adds that buffer to octo_buffers
+              OctoBuffer:new {
+                bufnr = bufnr,
+                number = obj.number,
+                repo = entry.repo,
+                node = obj,
+              }
+
               writers.write_title(bufnr, obj.title, 1)
               writers.write_details(bufnr, obj)
               writers.write_body(bufnr, obj)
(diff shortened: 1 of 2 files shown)


lua/octo/pickers/telescope/previewers.lua
```diff
@@ -38,6 +38,17 @@ local issue = defaulter(function(opts)
...
+              -- We have to create a new buffer since writers.write_state relies on
+              -- octo_buffers containing a buffer at a index of the current buffer number.
+              -- OctoBuffer:new adds that buffer to octo_buffers
+              OctoBuffer:new {
+                bufnr = bufnr,
+                number = obj.number,
+                repo = entry.repo,
+                node = obj,
+              }
+
```
pwntester, line 51: Wondering if it would be better to make `write_state` take an optional `isDraft` boolean to avoid creating (potentially) hundreds of OctoBuffers. Alternatively we could show if the PR is in draft or not, in a new Telescope/fzf-lua column (eg: https://github.com/pwntester/octo.nvim/blob/master/lua/octo/pickers/telescope/entry_maker.lua#L8-L41) so it does not need to be showed in the preview buffer. Thoughts?

HampusMat: It would definetly be better to show if PRs are drafts in a picker column. That would turn this bugfix PR into a feature PR though but thats maybe okay

pwntester: A new one would be better. Thanks!

HampusMat:
Closing since #485 will make including `[DRAFT]` in previews unnecessary.

PR #499, opened 2024-02-16
Allowing configuration of `gh` binary path

samjwillis97:
This allows defining the path of the `gh` binary

### Describe how you did it
Following the convention set out by other configurable values I simply added in `gh_cmd` with a default of `"gh"`.
For all uses of `command = "gh"` I changed it to `command = config.values.gh_cmd`

### Describe how to verify it
Setting the `gh_cmd` value in setup to the path of a working Github CLI binary

### Special notes for reviews

If this isn't something you are interested in including that is all good. I made the changes so I could use this plugin with my neovim configuration and thought it may come in handy to others as well 😄


pwntester:
Thanks!

Issue #519, opened 2024-03-17
Error execute vim.schedule: `attempt to index field defaultBranchRef` when do `octo repo list`, picker fzf-lua


MadKuntilanak:
## Issue Description

Type: *bug report* 

### Describe what happened 
Error when do `:Octo repo list` with picker fzf-lua (but okay with telescope)

```bash
Error executing vim.schedule lua callback: ....local/share/nvim/lazy/octo.nvim/lua/octo/ui/writers.lua:89: attempt to index field 'defaultBranchRef' (a userdata value)
stack traceback:
	....local/share/nvim/lazy/octo.nvim/lua/octo/ui/writers.lua:89: in function 'write_repo'
	...share/nvim/lazy/octo.nvim/lua/octo/model/octo-buffer.lua:89: in function 'render_repo'
	...m/lazy/octo.nvim/lua/octo/pickers/fzf-lua/previewers.lua:318: in function 'cb'
	...0x/.local/share/nvim/lazy/octo.nvim/lua/octo/gh/init.lua:160: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```

### How to reproduce it (as minimally and precisely as possible)

1. `:Octo repo list`
2. Error

### Tell us your environment
nvim nightly latest

milogert:
I think the problem is that you have a repository that doesn't have any default branch set. I'll work on fixing this though, thanks!

PR #521, opened 2024-03-20
fix: lack of default branch name causes repo display issue

milogert:

When a repo doesn't have a default branch name (i.e. a totally empty repo) then Octo gives an error when trying to view that repo.

This fixes that. Additionally, because when using the fzf-lua picker since the repo is previewed the error can show up when that repository is highlighted.

Fixes https://github.com/pwntester/octo.nvim/issues/519

### Describe how you did it

Check for the default branch ref and if it's `vim.NIL` the name just becomes nil. This will prevent displaying the name as well as fetching the README from the default branch.

### Describe how to verify it

Make a repo on Github and don't configure anything (no README, no .gitignore, no license file, etc.), then run `:Octo repo view you/your-repo`. Make sure there is no error as seen in #519 

### Special notes for reviews

N/A

### DIFF
--- a/lua/octo/ui/writers.lua
+++ b/lua/octo/ui/writers.lua
@@ -86,7 +86,13 @@ function M.write_repo(bufnr, repo)
 
   add_details_line(details, "Name", repo.nameWithOwner)
   add_details_line(details, "Description", repo.description)
-  add_details_line(details, "Default branch", repo.defaultBranchRef.name)
+  local defaultBranchRefName
+  if repo.defaultBranchRef == vim.NIL then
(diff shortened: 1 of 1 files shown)


milogert:
@MadKuntilanak would you be able to verify this one?

pwntester (approved):

MadKuntilanak:
> would you be able to verify this one?

yup i verified this with:
1. Create an empty folder and make repo with zero git suff
2. Open `octo repo list` or `ocro repo view <repo_link>`
3. Doing this doing that, success, no errors

Thanks!

PR #538, opened 2024-04-20
Feature: LSP navigation updates diffs

### LINKED ISSUE #535: When using `use_local_fs = true`, go to definition doesn't open a diff buffer
## Issue Description

Type: *feature request*

When I am using the `use_local_fs` config option, the benefit is that I can use my local files and lsp to navigate naturally around the code. But when I use go to definition on a newly added function for example in another file, the new function opens up without the side-by-side diff information.

In practice this means I need to select the file from the `files changed` panel to open it with this info.

I'd like to provide screenshots but my example is with code I cannot share - let me know if you want me to set up a dummy example for you.

I would expect that lsp go to definition would open the same buffer that is opened when I manually select the file from `files changed` so that it shows the rich diff and I can navigate the whole PR using lsp

1. Create a PR with at least two files
2. Create a function or something else in one file that is referenced in the other file
3. Make sure you are using `use_local_fs = true`
4. Have an lsp that supports go to definition with your chosen language
5. Review the PR with Octo (Octo [url] && Octo review start)
6. Open the file with the reference and go to definition on it
7. See that there is no diff information in the buffer and that the left buffer doesn't change from the original file

### Tell us your environment
I'm using Go and this version of Octo:
`  "octo.nvim": { "branch": "master", "commit": "a511b52fbf040f928a24deb72e17ec465613b442" },`

nope :) I love Octo, this is the only feature that is missing to make me stop using the GitHub UI for PR reviews (that and multiline comments/suggestions)

Thanks!!

delabere:
When you navigate using LSP or a command like `gd` while reviewing a change using the `use_local_fs` option, you lose the diff, even if the file you navigated to was part of the change.

This change allows you to navigate around a PR's changed files using your LSP

https://github.com/pwntester/octo.nvim/assets/29151020/47f26fb0-3769-42c9-bfa0-4c76b756308f

Fixes #535 

### Describe how you did it
Review layouts keep track of `files`, these are the files that you see populated in the files pane after opening a review.
After entering a buffer, if the filepath is not `octo://...`, I assume that it is a local file and I check whether the file is part of the layout's `file`'s. If it is then I update the review to that file using `review.layout:set_file(file)`

### Describe how to verify it
You need to use `use_local_fs` and review a PR that has multiple files, where you can navigate between those files via LSP.

### Special notes for reviews
There is some unintended behaviour with this change. When you navigate to the right buffer in the diff you see this message:

image

It doesn't happen if I comment out `review.layout:set_file(file)` in my change, so I assume that the side-effects of this are triggering more autocomands


jondkinney:
Seems like this would be a good change! Will this be included in a future release sometime soon?

delabere:
@pwntester 👋 just giving this one a nudge

pwntester:
@delabere took a look some time ago and could not figure out why its popping that autocmd error. Will take another look since merging it in this state will results in issues complaining about the error.

kapral18:
I don't see no warnings in latest nvim.

delabere:
@kapral18 I don't think we would expect any merge conflicts because I added new code in my PR.

I've swapped this out to use the `layout.set_current_file` function and although it didn't see to work for you, it seems to be working great for me - do you mind checking it out to make sure it works for you as well?

@pwntester I'm not seeing those errors anymore either - would you be happy to get this one merged in?

kapral18:
hey @delabere I checked the changes but for some reason it doesn't work


https://github.com/user-attachments/assets/ed0a3c3b-7873-42dd-b03d-c93443963c29

kapral18:
Also looking at this feature seems like it could be a good opt-in option, like `follow_lsp_navigation`, we can make it on by default, but looks to me that it's nice to have an escape hatch. Because I can also see value in maintain the current behavior where on the left side you have a snapshot of the original diff. For example in scenarios where previously you had everything in one file and then it was distributed into separate child files. This way having a frame of reference can actually be helpful.

wdyt @delabere  @wd60622

delabere:
@kapral18 an option sounds good but I don't have strong feelings about it. Feel free to steal and build on my changes if you'd like to. Otherwise I can add that when I get some time.

Can you try testing with a minimal config? I wonder if you have anything else that might be interfering with it, like an existing auto command for example 🤔

Also worth double checking that you've pulled the latest commits from master of you're cloning it down

kapral18:
@delabere ok I think I know what's going on. If you `gd` into a file that's not part of `review.layout.files` it skips `set_current_file`

https://github.com/pwntester/octo.nvim/pull/538/files#diff-551b860db1f227d6d4d20dfc8a8a09669a68b474e514e85baf9f22c2b4a85e04R55

delabere:
That's the intended behaviour 👍

kapral18:
@delabere feel free to merge this one I will add my suggestion in a separate PR. @wd60622 can we merge this?

wd60622:
Merged it in! Nice work @delabere 
We can iterate if needed

vuthanhtung2412:
can some share with me their neovim config? I am using lazy vim and I couldn't get it to work :(((
[This](https://github.com/vuthanhtung2412/nvim-config/blob/94cb08a74165cdc824fde3b3b9ed73fc5943910b/lua/plugins/octo.lua#L1) my lua file for octo using LazyVim

Below is video that show my attempt interaction with LSP from octo.nvim

https://github.com/user-attachments/assets/f3456ba0-8b02-489a-913f-851d1477cc57

Thanks a lot in advance

PR #540, opened 2024-04-21
fix: "go to file" falls back to finding the file from the git root

### LINKED ISSUE #539: "go to file" function only looks at current directory, not git root
## Issue Description

"go to file" function only looks at current directory, not git root

Type: *bug report*

When pressing `gf` while on a comment tagged to a file/line in a PR, with the current working directory not being at the root of the git repository, an error is returned.

The file would be found and `gf` would take the user to the file.

1. `cd` into a subdirectory of a git repository
2. open a PR with comments that are tagged to files
3. press `gf` to try to go to one of the files

### Tell us your environment

MacOS, neovim 0.9.5, latest octo.nvim

PR incoming.

mogest:

At present, "go to file" expects the file to be located relative to the current working directory.  This commit still looks for the file at the current directory, but falls back to using the git root path.  This enables "go to file" when the user is at a path inside the repository.

Fixes #539

### Describe how you did it

As above, fallback to using the git root path to find files.

### Describe how to verify it

1. `cd` into a subdirectory of a git repository
2. Open a PR in Octo that has comments with tagged files
3. Press `gf` on one of those comments and it will now successfully take you to the tagged file

### DIFF
--- a/lua/octo/navigation.lua
+++ b/lua/octo/navigation.lua
@@ -58,6 +58,16 @@ function M.open_in_browser(kind, repo, number)
   pcall(vim.cmd, "silent !" .. cmd)
 end
 
+local function open_file_if_found(path, line)
+  local stat = vim.loop.fs_stat(path)
+  if stat and stat.type then
+    vim.cmd("e " .. path)
+    vim.api.nvim_win_set_cursor(0, { line, 0 })
+    return true
+  end
+  return false
+end
+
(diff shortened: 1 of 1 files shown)


pwntester (approved):

pwntester (approved):

pwntester:
Thanks for the PR and sorry for the late review!

PR #554, opened 2024-06-07
fix: load octo in the correct buffer

folke:

When Octo loads on `BufReadCmd octo://*`, it doesn't pass the correct buffer and can
as such eventually load in a different buffer

### Describe how you did it

Simply pass the event buffer and wrap the result callback in a `nvim_buf_call` to ensure we render in the correctbuffer

### Describe how to verify it

Add `blank` to `sessionoptions`.
Make a split with one normal file and with an octo buffer.
Make the normal file the current window.
Make a session
Exit Neovim
Open NEovim
Restore Session
The octo buffer will load in the correct window,
Without this PR, it would load in the file window.

### DIFF
--- a/lua/octo/autocmds.lua
+++ b/lua/octo/autocmds.lua
@@ -18,7 +18,6 @@ function M.setup()
     group = "octo_autocmds",
     pattern = { "octo://*" },
     callback = function(ev)
-      dd(ev)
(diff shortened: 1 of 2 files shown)


pwntester:
Thanks!

PR #591, opened 2024-09-04
Fix human readable in PR comments

### LINKED ISSUE #588: Human readable difference wrong on new comment in PRs
Suspect this has something to do with the UTC offset and the initial comment being at local time zone but then stored in UTC

williambdean:
Fixes the PR comments be created in local but stored in UTC time

Closes #588


pwntester (approved):

pwntester:
Thanks @wd60622!

PR #598, opened 2024-09-12
feat: allow disabling auto review thread opening

### LINKED ISSUE #536: Less jumping around due to folds when opening comment threads during review
## Issue Description

Type: *feature request*

When reviewing a PR, in the file diff view, I find it a bit hard to follow all the jumping around that folds are causing when moving cursor over a line which has a comment thread.  
The root cause for this is that the other window of the split (most often the left window in my case) is replaced by the comment thread view, so the diff view stops showing its colors + the folds are automatically open, causing a lot of visual noise IMO.

There is also the annoyance of the thread window going away if I move around the file, when sometimes I am writing/reading a comment and want to check something 20 lines above.

1. First, I would like an option to only trigger  comment thread display through a dedicated command, to prevent jumps when just moving around the file. Inside the comment thread buffer, `q` in normal mode would close the buffer and go back to diff view (this could/should be bindable to let users chose their keybinds)

2. To go even further, we could maybe make use of nvim's floating windows ? Something like lsp hover windows where hovering over a line with comments, to avoid covering up the other part of the diff. I am less sure about this.

I have a more or less working version of step 1 that disables the autocommand that shows threads and adds support for manually opening/closing the thread buffer. I could open a PR, I am opening this issue to gauge interest.

Cheers

GuillaumeLagrange:
Add a config option to disable the auto thread opening on cursor move during review

Due to how diff works in nvim, I found it very hard to focus whenever there was more than one or two comments on a file

Fixes #536

### Describe how you did it
- Add the config option
- Create a command to call show review thread so that users can bind it
- When the config option is set,
    - do not create the cursor move autocmd that calls show review thread
    - create an local buffer keybind to close the comment buffer on q since it no longer auto closes when going back to diff

### Describe how to verify it

Octo config
```lua
  opts = {
    reviews = {
      auto_show_threads = false,
    },
  },
```

Open a pr review
- Go on a thread, call `<CMD>Octo review thread<CR>`
or
- Create a comment using `<CMD>Octo comment add<CR>`

See buffer open, and press q to close it

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,9 @@ require"octo".setup({
       direction = "DESC"                   -- either DESC or ASC (https://docs.github.com/en/graphql/reference/enums#orderdirection)
     }
   },
+  reviews = {
+    auto_show_threads = true,              -- automatically show comment threads on cursor move
+  },
   pull_requests = {
     order_by = {                           -- criteria to sort the results of `Octo pr list`
       field = "CREATED_AT",                -- either COMMENTS, CREATED_AT or UPDATED_AT (https://docs.github.com/en/graphql/reference/enums#issueorderfield)
--- a/lua/octo/autocmds.lua
+++ b/lua/octo/autocmds.lua
@@ -1,4 +1,6 @@
 local vim = vim
+local config = require "octo.config"
+
 local create = vim.api.nvim_create_augroup
 local define = vim.api.nvim_create_autocmd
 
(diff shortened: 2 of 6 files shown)


pwntester:
Thanks @GuillaumeLagrange looks good to me and actually may use it myself 😄

GuillaumeLagrange:
I have noticed a small regression that I believe this causes: if you open a comment from the left diff window (comment buffer opens on the right window), the `diffthis` is not properly applied to the right window on closing the comment buffer.

I'll open a PR with the fix

GuillaumeLagrange:
I can't reproduce, it must have been a half baked local feature causing it, sorry for the useless ping

PR #606, opened 2024-09-13
fix(repo parsing): considers cases where git repo has a trailing /

### LINKED ISSUE #605: bug: Cannot find repo when repository cloned with a trailing /
## Issue Description

Type: *bug report*
### Octo can't find repo if cloned with a trailing '/'

When running an Octo command without specifying the repository, it fails to resolve it if repository was cloned with a trailing '/'

### How to reproduce it
1. `git clone https://github.com/pwntester/octo.nvim/ && cd octo.nvim && nvim`
2. `Octo issue list`

This is caused by line 152 of `./lua/octo/utils.lua`
```lua
function M.parse_remote_url(url, aliases)
  -- remove trailing ".git"
  url = string.gsub(url, ".git$", "")
  -- remove protocol scheme
  url = string.gsub(url, "^[^:]+://", "")
  -- remove user
  url = string.gsub(url, "^[^@]+@", "")
  -- if url contains two slashes
  local segments = vim.split(url, "/")
  local host, repo
  if #segments == 3 then
    host = segments[1]
    repo = segments[2] .. "/" .. segments[3]
  elseif #segments == 2 then
    local chunks = vim.split(url, ":")
    host = chunks[1]
    repo = chunks[#chunks]
  end

  if aliases[host] then
    host = aliases[host]
  end
  if not M.is_blank(host) and not M.is_blank(repo) then
    return {
      host = host,
      repo = repo,
    }
  end
end
```
When the git repository was cloned with a trailing '/', segments has a length of 4 which the code doesn't account for
Changing line 152 to: `if #segments == 3 or #segments == 4 then` or `if #segments > 2 then` solves the issue

Hashino:

the automatic repository finding (when none passed as args) wouldn't work when the return of `git remote -v` had a trailing '/'. this is reproducible by adding a trailing '/' when cloning a repository

parser would split the lines with '/' as a delimiter. the trailing '/' made the result be an array of length 4 which wasn't considered.

Fixes #605 

### Describe how you did it
added condition for when the split segments has a length of 4.

### Describe how to verify it
```bash
git clone https://github.com/pwntester/octo.nvim/ && cd octo.nvim && nvim -c 'Octo issue list'
```

### Special notes for reviews

additional checking that `segment[4] == nil` could be added to ensure this change fixes only this edge case

the opposite approach could be made and use `if #segments > 2` so it becomes the default behavior for more edge cases

### DIFF
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -149,7 +149,7 @@ function M.parse_remote_url(url, aliases)
   -- if url contains two slashes
   local segments = vim.split(url, "/")
   local host, repo
-  if #segments == 3 or (#segments == 4 and segments[4] == '') then
+  if #segments == 3 or (#segments == 4 and segments[4] == "") then
     host = segments[1]
     repo = segments[2] .. "/" .. segments[3]
   elseif #segments == 2 then
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -149,7 +149,7 @@ function M.parse_remote_url(url, aliases)
   -- if url contains two slashes
   local segments = vim.split(url, "/")
   local host, repo
-  if #segments == 3 then
+  if #segments == 3 or (#segments == 4 and segments[4] == '') then
     host = segments[1]
     repo = segments[2] .. "/" .. segments[3]
   elseif #segments == 2 then


lua/octo/utils.lua
```diff
@@ -149,7 +149,7 @@ function M.parse_remote_url(url, aliases)
   -- if url contains two slashes
   local segments = vim.split(url, "/")
   local host, repo
-  if #segments == 3 then
+  if #segments == 3 or #segments == 4 then
```
pwntester, line 152: Please add a check that the fourth segment is an empty string

pwntester (changes requested):

pwntester (approved):

pwntester:
Thanks @Hashino can you please fix the linting errors?

Hashino:
@pwntester done

Hashino:
@pwntester since this is my first contribution you'll need to approve the run of the workflows

pwntester (approved):

pwntester:
Thanks!

Issue #443, opened 2023-10-31
Feature Request: Develop Issue Command


arnold-c:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

I'm not sure if this is already implemented (I tried looking and couldn't find any documentation or mentions in the discussions/issues about it, but apologies if I just missed it), but I think it would be useful to be able to use the `gh issue develop #X --checkout` command to turn an issue into a new branch.

### Describe what you expected to happen

The GitHub CLI has an option to turn an issue into a branch and then check it out on your local machine (among other options).

https://cli.github.com/manual/gh_issue_develop

I think it would be useful to be able to either: 1) open up the Issue list, open an issue in a buffer, and then use an Octo command to be able to send the above command to develop an issue branch; or 2) use an Octo command to bring up a Telescope viewer where you can select an issue to develop, and then you are given a dropdown to select the extra flags you would like to apply (like using templates for issue creation). Option 2) would be cleaner and faster, in my opinion, but likely also be more work.

I would offer to help, but I'm afraid I am not familiar enough with scripting in Lua and Telescope, so I don't think I'd be of much help.

Thanks for your work putting out a great plugin.

**EDIT:** Embarrasingly, I've just seen this discussion https://github.com/pwntester/octo.nvim/discussions/392 right near the top, but I think the feature request is still valid as it potentially provides a mechanism through the CLI to implement this? Please close if you feel it is not required.

wd60622:
I think both options would be very useful. I've been playing around with the GitHub CLI commands required here and it seems like there might be some permission issues when trying to create a linked branched to the upstream repo from a fork.

wd60622:
Hi @arnold-c 

I have implemented this in #609 . 1 and 2 are both available for telescope. However, I haven't implemented 2 for other pickers

PR #609, opened 2024-09-19
Add `Octo issue develop` command

wd60622:
Adds `issue develop` command. 

If on issue buffer, the current issue is checked out and the branch is created.
If not, then a picker is shown to select an issue to checkout and create a branch.

NOTE: Only implemented for the telescope picker

Related to #443

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -66,13 +66,12 @@ function M.setup()
         local bufnr = vim.api.nvim_get_current_buf()
         local buffer = octo_buffers[bufnr]
 
-        if not buffer then
+        if buffer and buffer.kind and buffer.kind == "issue" then
+          utils.develop_issue(buffer.node.number)
+        else
           local opts = M.process_varargs(repo, ...)
           picker.issues(opts, true)
-          return
         end
-
-        utils.develop_issue(buffer.node.number)
       end,
       reopen = function()
         M.change_state "OPEN"
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -62,6 +62,18 @@ function M.setup()
       close = function()
         M.change_state "CLOSED"
       end,
+      develop = function(repo, ...)
+        local bufnr = vim.api.nvim_get_current_buf()
+        local buffer = octo_buffers[bufnr]
+
+        if not buffer then
+          local opts = M.process_varargs(repo, ...)
+          picker.issues(opts, true)
+          return
+        end
+
+        utils.develop_issue(buffer.node.number)
+      end,
       reopen = function()
         M.change_state "OPEN"
       end,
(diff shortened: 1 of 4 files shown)


lua/octo/commands.lua
```diff
@@ -62,6 +62,18 @@ function M.setup()
...
+      develop = function(repo, ...)
+        local bufnr = vim.api.nvim_get_current_buf()
+        local buffer = octo_buffers[bufnr]
+
+        if not buffer then
+          local opts = M.process_varargs(repo, ...)
+          picker.issues(opts, true)
+          return
+        end
+
```
wd60622, line 74: Need to add clause for not issue buffer

wd60622 (commented):
Still need to work through any documentation, mappings, or configuration

Any thoughts here?

pwntester:
Sorry, was not able to take a look. Will do it today

wd60622:
> Sorry, was not able to take a look. Will do it today

Thanks @pwntester 

Note: I adapted the `M.process_varargs` which allows for repo as argument. This potentially shouldn't work for develop as the issue numbers from another repo wouldn't make sense

lua/octo/commands.lua
```diff
@@ -62,6 +62,17 @@ function M.setup()
...
       end,
+      develop = function(repo, ...)
+        local bufnr = vim.api.nvim_get_current_buf()
+        local buffer = octo_buffers[bufnr]
+
+        if buffer and buffer.kind and buffer.kind == "issue" then
+          utils.develop_issue(buffer.node.number)
+        else
+          local opts = M.process_varargs(repo, ...)
+          picker.issues(opts, true)
```
pwntester, line 73: Do you expect to pass any of the arguments handled by varargs to `Octo issue develop`? If so, this should be ok

pwntester (approved):

lua/octo/commands.lua
wd60622: Do not expect to pass any args here. Was using the functions that already exist in order to get the current repo (when repo is not passed). But could just duplicate the logic instead. 

This works as is but could cause some confusion if digging into this.

What are your thoughts?

wd60622:
TODO: 

- [ ] Add command to the README

Any other items?

lua/octo/commands.lua
pwntester: I think we can pass `repo` and `develop` as part of the opts so there should not be any need for handling varargs or passing develop as a separate argument.

wd60622: Are you suggesting a change then?

wd60622: The issue with passing `repo` as an argument is that the location of that repo locally is not known. So, only the current repo is truely supported.

wd60622: I can create an issue to clean up this in the future

PR #614, opened 2024-09-25
fix: remove escape_char since we are now using """ quoted strings

### LINKED ISSUE #613: Quotes are being escaped
After saving an issue, the \" character is being escaped

Testing with '

Seems to only be happening with the \" character

Using `gh` version 2.57.0

pwntester:

Closes #613

### Special notes for reviews

### DIFF
--- a/lua/octo/gh/graphql.lua
+++ b/lua/octo/gh/graphql.lua
@@ -3229,14 +3229,6 @@ query {
 }
 ]]
 
-local function escape_char(string)
-  local escaped, _ = string.gsub(string, '["\\]', {
-    ['"'] = '\\"',
-    ["\\"] = "\\\\",
-  })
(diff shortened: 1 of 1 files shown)


wd60622:
I am having that this works for two double quotes but I am getting an error from `gh` CLI when doing a single double quote.

pwntester:
You are right, I think the problem is to have a double quote at the end of the comment body since it gets followed by the closing `"""` double quotes and it will unbalance the quotes. However, that does not happen when you add a leading `"` and no trailing one 🤔

wd60622 (approved):
Working for me. Just has the issue when one `\"` is at the end of the comment block. But that shouldn't be a blocker

wd60622:
This one good to merge?

Issue #611, opened 2024-09-21
Error with picker fzf-lua


dpetka2001:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)
When using `fzf-lua` as picker, I get an error when doing `Octo search`. 
```lua
Error executing vim.schedule lua callback: ...v/lazy/octo.nvim/lua/octo/pickers/fzf-lua/previewers.lua:115: bad argument #1 to 'tonumber' (value expected)
stack traceback:
	[C]: in function 'tonumber'
	...v/lazy/octo.nvim/lua/octo/pickers/fzf-lua/previewers.lua:115: in function 'populate_preview_buf'
	...azyVimDev/lazy/fzf-lua/lua/fzf-lua/previewer/builtin.lua:263: in function 'populate_preview_buf'
	...azyVimDev/lazy/fzf-lua/lua/fzf-lua/previewer/builtin.lua:280: in function ''
	vim/_editor.lua: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```
With `telescope` picker it works without problems.

### Describe what you expected to happen
Expect to work without errors, just like with `telescope` picker.

### How to reproduce it (as minimally and precisely as possible)

1. Just do `Octo search`

### Tell us your environment
LazyVim distro on Linux Mint 21.3 with latest kitty terminal and latest Neovim nightly.

PR #619, opened 2024-09-28
fix(fzf-lua): when no search result

phanen:
Fzf-lua previewer didn't handle error when no matched result found.

Fix https://github.com/pwntester/octo.nvim/issues/611.

### Describe how you did it
Do noting when no matched result found.


GuillaumeLagrange:
Hey, the linter is failing, could you fix this?

phanen:
Fixed.

dpetka2001:
Would be great if this could be merged. I'm using this PR locally and haven't observed any problems thus far. It would be great for users that use `fzf-lua` instead of Telescope.

PR #546, opened 2024-05-23
Improve the `Octo pr list` command

alexkalderimis:
### Describe what this PR does / why we need it

This PR includes improvements to the `pr list` command. Specifically:

- it changes the fetch implementation to one based on REST rather than GraphQL, since GraphQL lacks several important features, such as the ability to filter on authors or assignees or labels.
- it abandons the attempt to fetch all PRs via pagination, since repos with hundreds or even thousands of PRs will cause this to be inefficient or unusable
- it changes the display in telescope of PRs to show authors and branches, and allows the user to filter based on these values. This is very important in organizations where practices such as including ticket IDs in branch names are followed.

### Does this pull request fix one issue?

NONE

### Describe how you did it

The main changes are in `pickers/telescope/entry_maker.lua` and `pickers/telescope/provider.lua`. These files define both the mechanism for retrieving pull requests and how they are passed to the telescope picker.

A new function `gen_from_pull_request` is added in `entry_maker.lua` to handle the display of the pull request data, including the addition of the new author and branch columns. The `ordinal` for pull requests now includes the author and branch information, allowing this to be used in filtering (displaying this is not enough).

The retrieval is changed in `telescope/provider.lua` (specifically in the `pull_requests` function). Here we call `gh pr list` rather than using the exisiting GraphQL implementation in order to provide access to the additional features that allows.

The retrieved records are modified to conform to the GraphQL result schema, meaning that they can be used in existing functions if necessary.

### Describe how to verify it

The following actions will be useful in verification:

- run `Octo pr list` and then use telescope to filter by author name
- run `Octo pr list author=@me` and see that only your PRs are shown

PR #635, opened 2024-10-14
pickable branches

legobeat:
- **Show branch names in the PR picker**
  

- **Allow PRs to be filtered by branch and author**
  

- **Replace PR list implementation**
  Prior to this change, we were using the GraphQL resources to retrieve
  pull requests. This is limited, since there are a number of very useful
  queries that are not available in GraphQL, including filtering PRs by
  author or assignee.
  
  This change replaces the GraphQL implementation with one based on
  `gh pr list`, which uses the REST API, and is more functional.
  
  This allows filtering by author, assignee, branches (both head and
  base), state and labels, as well as arbitrary search strings.
  
  To handle large repos gracefully, we abandon the attempt to retrieve all
  pull requests by paginating, since this is potentially unbounded.
  Instead we fetch at most a configurable limit of pull requests, and
  expect the user to use filtering on the list to improve relevance.
  

- **Update help text**
  

- **fix: handle nil pull.author.id**
  

- **fix: --paginate**
  

- **fix: validate & document pull_requests.limit**
  default 100
  

- **feat: use pull_requests.limit in fzf provider**


macovsky:
@legobeat thanks so much for this!

lua/octo/pickers/fzf-lua/pickers/prs.lua
```diff
@@ -41,6 +41,8 @@ return function(opts)
         "api",
         "graphql",
         "--paginate",
```
wd60622, line 43: I have this causing an error saying it is not one of the available flags

legobeat: Good catch! Reverted this part.

README.md
```diff
@@ -144,6 +144,7 @@ require"octo".setup({
       field = "CREATED_AT",                -- either COMMENTS, CREATED_AT or UPDATED_AT (https://docs.github.com/en/graphql/reference/enums#issueorderfield)
       direction = "DESC"                   -- either DESC or ASC (https://docs.github.com/en/graphql/reference/enums#orderdirection)
     },
+    limit = 100                            -- maximum number of pull requests in `Oto pr list` results
```
elsesiy, line 147:
```suggestion
    limit = 100                            -- maximum number of pull requests in `Octo pr list` results
```

wd60622:
This doesn't work for me still. I commented out the "--paginate" as it isn't a flag in the command used. I prefer the previous entry_maker a bit more. Plus with the colors & symbols here: https://github.com/pwntester/octo.nvim/pull/637

wd60622:
Also noting that this doesn't impact the `Octo pr search command`. Is that as intended?

wd60622:
I think much of the original requirements can be done with the current `Octo pr search` and being combined with the advanced search functionality: https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests

If the goal is to have it in the fuzzy finder and move away from GitHub search then that is another item. 

What do people think?

macovsky:
> I think much of the original requirements can be done with the current `Octo pr search` and being combined with the advanced search functionality: https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests

one slight problem with `Octo pr search` is that one can't merge/checkout PRs right in telescope with picker bindings (these bindings work for me only in `Octo pr list`)

> If the goal is to have it in the fuzzy finder and move away from GitHub search then that is another item.
> 
> What do people think?

IMHO it's nice to have branch/author in fuzzy finder

wd60622:
> IMHO it's nice to have branch/author in fuzzy finder

I guess I find the picker a bit cumbersome in the current form. How about adding the formatting of the user and the branch as done in the PR / issue details in order to provide some separation of the row information.

> one slight problem with `Octo pr search` is that one can't merge/checkout PRs right in telescope with picker bindings (these bindings work for me only in `Octo pr list`)

Adding those bindings could be a good addition and easier solution for this problem.

macovsky:
> > one slight problem with `Octo pr search` is that one can't merge/checkout PRs right in telescope with picker bindings (these bindings work for me only in `Octo pr list`)
> 
> Adding those bindings could be a good addition and easier solution for this problem.
> 
> That would be done here: https://github.com/wd60622/octo.nvim/blob/88d9fe8e90fa3f5fdeedf4d8d7c1b1abf27f825f/lua/octo/pickers/telescope/provider.lua#L558-L559
> 
> in the same manner as:
> 
> https://github.com/wd60622/octo.nvim/blob/88d9fe8e90fa3f5fdeedf4d8d7c1b1abf27f825f/lua/octo/pickers/telescope/provider.lua#L341-L344
> 
> But ideally, the user would be able to configure these a bit more.

do you mind having a look at https://github.com/pwntester/octo.nvim/pull/647?

legobeat:
Apologies for dropping a half-baked version without timely follow-up!

> This doesn't work for me still. I commented out the "--paginate" as it isn't a flag in the command used.

Thanks. I have rebased on `master` now, fixed that oversight and done some manual testing now.

> I also prefer the previous entry_maker a bit more. Plus with the colors & symbols here: https://github.com/pwntester/octo.nvim/pull/637

Nice!

> Also noting that this doesn't impact the Octo pr search command. Is that as intended?

There are some tradeoffs with each approach (see description on #546 for some motivations). One one hand I'm not sure too what extent it makes sense to have separate `search`/`list` commands with differing behavior and implementation but on the other it _is_ convenient to use it as differentiator between calling graphql or not, as either can be desirable and it would be a chore to have to configure desired default for each repo... So maybe a single unified implementation which can be configured, but still two differing commands with differing defaults is a better approach...?

It looks like there are some snags/bugs (different ones!) in both implementations at this point. Even if existing bugs are fixed I still have situations where this `pr list` approach is preferred. Open to suggestions on how this PR can be improved in that sense.

> How about adding the formatting of the user and the branch as done in the PR / issue details in order to provide some separation of the row information.

I think that sounds like a great idea regardless of all the above! That would solve the motivation of having author name viewable and filterable.  Though I guess that would mean that field/column-based sorting would become quite cumbersome?

doc/octo.txt
```diff
@@ -59,7 +59,7 @@ See |octo-command-examples| for examples.
                           <C-b>   Open PR in browser.
                           <C-o>   Checkout PR.
                           <C-y>   Copy URL to system clipboard.
-                        Available filter keys: see <https://docs.github.com/en/graphql/reference/input-objects#issuefilters>
+                        Available filter keys: author,assignee,label,state,head,base,search,limit
```
pwntester, line 62:
```suggestion
                        Available filter keys: author, assignee, label, state, head, base, search, limit
```

pwntester (approved):

pwntester:
Thanks for taking care of this PR @legobeat. It is still on draft state, are you planning on doing any more work on this PR?

pwntester:
Thanks for the update @legobeat We have created a discussion room in Matrix in case you want to join us https://matrix.to/#/#octo.nvim:matrix.org

Also, happy to make you a collaborator if you are up for it

PR #643, opened 2024-10-21
Add discussion list support

wd60622:
Ability to list discussions for a repository

Adds the `discussion list` command 

Related to #239

### DIFF
--- a/lua/octo/pickers/telescope/previewers.lua
+++ b/lua/octo/pickers/telescope/previewers.lua
@@ -8,6 +8,15 @@ local pv_utils = require "telescope.previewers.utils"
 local ts_utils = require "telescope.utils"
 local defaulter = ts_utils.make_default_callable
 
+local discussion = defaulter(function(opts)
+  return previewers.new_buffer_previewer {
+    title = opts.preview_title,
+    define_preview = function(self, entry)
+      return self.state.bufnr, self.state.bufname
+    end,
+  }
+end)
+
 local issue = defaulter(function(opts)
   return previewers.new_buffer_previewer {
     title = opts.preview_title,
@@ -174,6 +183,7 @@ local issue_template = defaulter(function(opts)
 end, {})
 
 return {
+  discussion = discussion,
   issue = issue,
   gist = gist,
   commit = commit,
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -52,6 +52,12 @@ function M.setup()
     search = function(...)
       M.search(...)
     end,
+    discussion = {
+      list = function(repo, ...)
+        local opts = M.process_varargs(repo, ...)
+        picker.discussions(opts)
+      end,
+    },
     issue = {
       create = function(repo)
         M.create_issue(repo)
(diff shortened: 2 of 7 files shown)


wd60622:
`:Octo discussion list` command

discussion-list

Currently there is not a buffer that supports interacting with discussion but user can view in the previewer and <C-y> for URL

pwntester:
Thanks for adding support for discussions @wd60622 Do you want to merge this in before adding a discussion type of buffer? Im ok with showing the preview only and maybe opening it a browser on `<CR>` for now (until a dedicated octo buffer type is created)

wd60622:
Thats my idea as well. i can break it down a bit more a create issues for it. Currently the <CR> copies the number to register. C-y works for url. I havent figured out the how to open discussion in browser just yet. Is that desired <CR> behavior?

pwntester:
From a UX point of view, I think users may expect to "enter" the discussion when pressing CR. If we cant show it in a buffer for now, I think the behaviour that would make more sense is to open it in a browser. You can use https://github.com/pwntester/octo.nvim/blob/master/lua/octo/navigation.lua#L12-L23 for it

wd60622:
> From a UX point of view, I think users may expect to "enter" the discussion when pressing CR. If we cant show it in a buffer for now, I think the behaviour that would make more sense is to open it in a browser. You can use https://github.com/pwntester/octo.nvim/blob/master/lua/octo/navigation.lua#L12-L23 for it

Great. I will get to it when I can!

lua/octo/ui/writers.lua
```diff
@@ -76,6 +76,109 @@ local function add_details_line(details, label, value, kind)
...
+  local details = {}
+
+  -- clear namespace and set vt
+  vim.api.nvim_buf_clear_namespace(bufnr, constants.OCTO_REACTIONS_VT_NS, 0, -1)
+
+  -- print("This is the discussion")
+  -- vim.print(discussion)
+
+  -- add_details_line(details, "Number", discussion.number)
+  --
```
wd60622, line 89:
```suggestion
```

lua/octo/pickers/telescope/previewers.lua
```diff
@@ -8,6 +8,58 @@ local pv_utils = require "telescope.previewers.utils"
...
+            local result = vim.fn.json_decode(output)
+            local obj = result.data.repository.discussion
+
+            writers.write_title(bufnr, tostring(obj.title), 1)
+            writers.write_discussion_details(bufnr, obj)
+            writers.write_body(bufnr, obj, 11)
+
+            -- local line = vim.api.nvim_buf_line_count(bufnr) + 2
+            -- writers.write_upvotes(bufnr, obj, line)
+            -- writers.write_reactions(bufnr, obj.reactionGroups, line)
```
wd60622, line 48:
```suggestion
```

wd60622:
Hi @pwntester, thank you for the review. I have added the open in browser functionality. Please take a look and see if it works on your end.  I will begin to make some issues for the next steps.

PR #654, opened 2024-10-30
use "comment" over "comments" in review buffer

### LINKED ISSUE #648: Use comment over comments in review buffer
1 comments shows up in the review buffer which is grammatically incorrect.

macovsky:

Displays "comment" in review buffer instead of "comments"

Fixes #648

### Describe how to verify it

Make a comment in your review and see that it's displayed as "1 comment" in diff

### Checklist

### DIFF
--- a/lua/octo/reviews/file-entry.lua
+++ b/lua/octo/reviews/file-entry.lua
@@ -382,7 +382,9 @@ function FileEntry:place_signs()
 
           -- place the virtual text only on first line
           local last_date = comment.lastEditedAt ~= vim.NIL and comment.lastEditedAt or comment.createdAt
-          local vt_msg = string.format("    %d comments (%s)", #thread.comments.nodes, utils.format_date(last_date))
+          local comments_count = #thread.comments.nodes
+          local comments_word = comments_count == 1 and "comment" or "comments"
+          local vt_msg = string.format("    %d %s (%s)", comments_count, comments_word, utils.format_date(last_date))
           --vim.api.nvim_buf_set_virtual_text(split.bufnr, -1, startLine - 1, { { vt_msg, "Comment" } }, {})
           local opts = {
             virt_text = { { vt_msg, "Comment" } },


lua/octo/reviews/file-entry.lua
```diff
@@ -382,7 +382,9 @@ function FileEntry:place_signs()
 
           -- place the virtual text only on first line
           local last_date = comment.lastEditedAt ~= vim.NIL and comment.lastEditedAt or comment.createdAt
-          local vt_msg = string.format("    %d comments (%s)", #thread.comments.nodes, utils.format_date(last_date))
+          local comments_count = #thread.comments.nodes
+          local comments_word = comments_count == 1 and "comment" or "comments"
```
wd60622, line 386: Test multiline comment with branch checked out

lua/octo/reviews/file-entry.lua
```diff
@@ -382,7 +382,9 @@ function FileEntry:place_signs()
 
           -- place the virtual text only on first line
           local last_date = comment.lastEditedAt ~= vim.NIL and comment.lastEditedAt or comment.createdAt
-          local vt_msg = string.format("    %d comments (%s)", #thread.comments.nodes, utils.format_date(last_date))
```
wd60622, line 385: Test comment over here

wd60622 (commented):
I think this seems good. I checked out the branch and had the \"comments\" still when I added the comment but \"comment\" after loaded. I tried again and it wasn't like that.

Do you know if there is a different function at comment creation and rendering?

macovsky:
> I think this seems good. I checked out the branch and had the \"comments\" still when I added the comment but \"comment\" after loaded. I tried again and it wasn't like that.

hm, I tried multiple times and couldn't see "comments" anywhere, it just appears as "comment" for me right from the start. did you see it as "1 comments" or maybe even "0 comments"?

> Do you know if there is a different function at comment creation and rendering?

idk really, just looked in the code and couldn't find anything related

wd60622 (approved):
Makes sense. Thanks for taking a look. Lets push it through!

PR #661, opened 2024-11-02
Add name in user queries

wd60622:
This adds the name in parenthesis to the list of users in various commands.
Helps with assigning people's whose logins are not the same as their names and you know name over username

users


pwntester (approved):

pwntester:
Thanks for the Pull Request Will

PR #662, opened 2024-11-03
fix(pickers/telescope): handle nil selection in open

legobeat:

It may happen that `selection` is `nil`, which results in an error being logged.

### Describe how you did it

This silently skips the failing function in such cases.

### Describe how to verify it

Run `Octo pr search` and fiddle around with filtering and navigation

### Checklist


pwntester (approved):

pwntester:
Thanks!

PR #665, opened 2024-11-03
feat: match patterns in ssh_aliases

legobeat:

Some people may use "dynamic" remote hosts like `github.com-foo`, `github.com-bar`, etc. Having to additionally add each as an explicit `ssh_aliases` entry can become a chore.

### Describe how you did it

Interpret keys in `ssh_alieses` configuration as [Lua patterns](https://lua.org/pil/20.1.html). This allows an alias like `github\.com-.*`.

### Checklist


pwntester (approved):

btkramm:
This pull request created a bug when using `ssh_aliases`!

The issue is with the pattern in the `gsub` call. When you use `"^" .. alias .. "$"`, Lua treats special characters in `alias` literally, but you need to escape them to match them literally in the pattern.

In my case, `github.com-work` contains a hyphen (-) which has special meaning in Lua pattern, and the `gsub` call is unable to replace the host.

PR #670, opened 2024-11-09
pr display: fix collecting review states per reviewer

macovsky:

before this PR if reviewer commented first and then approved PR, he will still be displayed with "commented" state in "Reviewers" section because review states weren't collected

NONE

### Describe how to verify it

now if you comment/request changes and then approve PR, you'll be displayed with "approved" state

### Checklist

### DIFF
--- a/lua/octo/ui/writers.lua
+++ b/lua/octo/ui/writers.lua
@@ -380,19 +380,15 @@ function M.write_details(bufnr, issue, update)
     -- reviewers
     local reviewers = {}
     local collect_reviewer = function(name, state)
-      --if vim.g.octo_viewer ~= name then
       if not reviewers[name] then
-        if not reviewers[name] then
-          reviewers[name] = { state }
-        else
-          local states = reviewers[name]
-          if not vim.tbl_contains(states, state) then
-            table.insert(states, state)
-          end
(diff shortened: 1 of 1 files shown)


macovsky:
there is still one issue left though, the whole "strongest review state" thing doesn't reflect current review state per reviewer as github itself does. now if you

macovsky:
> there is still one issue left though, the whole "strongest review state" thing doesn't reflect current review state per reviewer as github itself does: if reviewer commented/requested changes and then author resolves comments and re-request the review github diplays reviewer with "requested" state but octo simply shows "strongest" state and ignores consequent review requests (that's a slight problem because I don't see if I need to re-request review or not).
> 
> there's simply no notion of "latest request": https://github.com/macovsky/octo.nvim/blob/e3b05a2873c324c1a42324d9162df93d8c15826f/lua/octo/utils.lua#L140-L152
> 
> what are your thoughts on this? IMO we would want to either match github and display the latest state (because octo is _for github_ anyways) or add the latest state to the "strongest state", so if reviewer commented/requested changes first and then was re-requested to review we would display two states close to reviewer: the first one and then pending.

we can move this discussion into a separate issue just in case. what do you think @wd60622?

PR #681, opened 2024-11-14
use <localleader> instead of <leader>

Groz17:
From `:h localleader`: `<LocalLeader>` is to be used for mappings which are local to a buffer.

NONE
### Describe how you did it
Simple substitution: `s/<leader>/<localleader>/g`

### Checklist


GuillaumeLagrange (approved):
Thanks for the PR! Lgtm

nicos68:
Hi. This breaks my configuration. The octo mappings starting with <localleader> do nothing now. Am I missing something ?
You can see my nvim config [here](https://codeberg.org/nise/dotfiles/src/branch/main/.config/nvim)

Groz17:
In `~/.config/nvim/lua/nise/keymaps.lua` or wherever you want, add `vim.g.maplocalleader = <char>`, replacing `<char>` with whatever character you want. I personally use comma but if you want to use space like before just type `vim.g.maplocalleader = " "`

nicos68:
OK thanks, never knew about this feature before. Maybe a word about that in the read me would be useful.

danrasmuson:
I'm glad this was added an it does seem like an improvement. However it did take a bit to figure out what happened.

nicos68:
I think some words about that in the readme would be enough.

danrasmuson:
@GuillaumeLagrange the way LazyVim does this is that have this changelog feature that comes up when I open neovim for the first time and it describes breaking changes and what not. I'm not aware if there is a standardized method for opting into or setting up this type of behavior in a changelog but I'd thought I'd bring it to awareness as once setup it would be helpful.

CleanShot 2024-11-21 at 13 47 07@2x

Also I understand that this type of system may take some energy to setup and might not be worth it right now, but good to think about these things even if we don't action on it yet.

danrasmuson:
@GuillaumeLagrange I learned today that putting [`!` at the end of a commit message marks the commit as including a breaking change](https://www.conventionalcommits.org/en/v1.0.0/?utm_source=chatgpt.com#commit-message-with--to-draw-attention-to-breaking-change). As you can see in the below screenshot when I updated my plugins with LazyVim it moved the updates with breaking commits to the top so that I can I beware of this. It knew this because the commit of the plugin update included `!`. This could be would be a good standard to include while going forward, especially if there are any especially taxing changes.

CleanShot 2024-12-04 at 16 19 58

PR #686, opened 2024-11-24
feat(notifications): show and read unread notifications

### LINKED ISSUE #119: Feature request - open issues and PRs via notifications
Starting to use the plugin more and more in my professional work I realize a gap between the WebUI and using Octo in NeoVim. I read notifications in the browser, click on them to do a requested review for example and then realize I could/want to actually do this with Octo. I need to  switch to the repository, run NeoVim, find the PR again with Octo and then start to review.
That's a cumbersome workflow which Octo does not benefit from. But I think we can easily start to narrow this gap.
The first link in the chain are the notifications for me. Sorry, confusing terminology. The notification to the user that there is a new notification objects (new entry in a thread object). Personally I have a tool that sends me desktop notification for this. I also use the mobile app with push notifications. Other user have e-mail messages for that. After all I think this isn't a topic Octo should try to solve. Maybe this is a thing for another little plugin that uses the new NeoVim notification API function. 
The second link is to open the list of open notifications, and three, click on it to interact with the threads subject object (e.g. an issue). And I think Octo can perfectly do this. We can query the API to get a user's notification. And then as for all other objects, we offer the user the list in Telescope. And when he selects a notification we open the related subject object and BAM, he gap got closed. :laughing:

One technical challenge is that unfortunately the notifications are exposed via the REST API endpoint, but not the GraphQL one. While this is not a problem by itself as the cli tool can also just query the REST endpoint, it brings some chaos into the code. We should consider to improve the plugins architecture first to push the dependencies to the borders and establish the Single-Responsibility-Principle (often misinterpreted by the name, don't get it wrong).
Another challenge would be the Telescope previewer as it would need to handle different objects instead of single specific one as it is done now. But again, I think this points us to great improvement in the plugins architecture. Btw: we should filter notifications only for those who refer to subject objects we support (i.e. issues, PRs, gists and commits, no releases or discussions).
Finally in the long long perspective we could think about how to render the specific comment in an issue/PR/gist that the thread notification is referring to. Thereby I mean showing it in the Telescope previewer as well as jumping/scrolling to it when a notification get opened.

All in all I think this whole feature isn't that big and not so hard to implement. Ignoring possible proceeding refactorings. But the effect and usability gain of the plugin is enormous. From my point of view...

tris203:
Add ability to see unread notifications, review and mark as read from within octo

image

Fixes #119 

### Describe how you did it
Use REST API for notifications as GraphQL is lacking

### Describe how to verify it

1. `:Octo notification` 
2. See unread notifications
3. scroll the list of PRs/Issues
4. Enter will open that issue/PR
5. `<leader>nr` will mark the notification as read
### Special notes for reviews
Currently limited to unread notifications only due to API limitations
Icons are in place for unread, for extension in the future

### Checklist


wd60622:
My telescope picker from the `Octo`  is pretty tight. Can we make the first column width based on the width of the longest word (maybe with some padding)

notifications

wd60622:
I tend to have use Octo from within the repo, how about a configuration to filter to just the notifications for the current repo. Then no need to show the repo name as well. What are your thoughts?

lua/octo/pickers/telescope/entry_maker.lua
```diff
@@ -694,6 +704,64 @@ function M.gen_from_octo_actions()
...
+
+    local displayer = entry_display.create {
+      separator = " ",
+      items = {
+        { width = 2 },
+        { width = 12 },
+        { width = math.min(#entry.obj.repository.full_name, 50) },
+        { width = math.min(#entry.obj.subject.title, 100) },
+      },
+    }
```
wd60622, line 728: The other pickers (pr list, issue list, discussion list, and search) use columns:  

- number
- Symbol
- title

Would it make sense to consolidate to that ordering? Maybe squeeze the repo before title

tris203:
> I tend to have use Octo from within the single repo, how about a configuration to filter to just the notifications for the current repo. Then no need to show the repo name as well. What are your thoughts? 

I don't think that particularly makes sense given the current API limitations. Given that we only show unread notifications, you would end up looking at an empty list 99% of the time

wd60622:
> > I tend to have use Octo from within the single repo, how about a configuration to filter to just the notifications for the current repo. Then no need to show the repo name as well. What are your thoughts? 
> 
> I don't think that particularly makes sense given the current API limitations. Given that we only show unread notifications, you would end up looking at an empty list 99% of the time. 
> 
> Also a general, what do I have to do would mean pinging around all over the place to see where you _might_ have notifications

I was thinking pull all notifications but filter before passing to picker

tris203:
> > > I tend to have use Octo from within the single repo, how about a configuration to filter to just the notifications for the current repo. Then no need to show the repo name as well. What are your thoughts?
> > 
> > 
> > I don't think that particularly makes sense given the current API limitations. Given that we only show unread notifications, you would end up looking at an empty list 99% of the time.
> > Also a general, what do I have to do would mean pinging around all over the place to see where you _might_ have notifications
> 
> I was thinking pull all notifications but filter before passing to picker

That works, but I don't know if with the telescope API if we can "mark an item read" without refreshing the whole list?

lua/octo/pickers/telescope/entry_maker.lua
tris203: im also not sure if that order makes sense for the notifications...just from a UI perspective

wd60622:
> That works, but I don't know if with the telescope API if we can "mark an item read" without refreshing the whole list?

But that affects even the case without the filter, right?

tris203:
> > That works, but I don't know if with the telescope API if we can "mark an item read" without refreshing the whole list?
> 
> But that affects even the case without the filter, right?

No, because we just grab the unread items list, and marking it read fires the API even and optimistically removes it from the selector without the need to refresh the whole thing.

We could change it to a `notification list` and a `notification list all` or something similar, but I don't know how that best fits with the command list, and your still going to end up looking at an empty list most of the time.

wd60622:
> No, because we just grab the unread items list, and marking it read fires the API even and optimistically removes it from the selector without the need to refresh the whole thing.

Seems like the same logic for the case where the list is already filtered

> We could change it to a `notification list` and a `notification list all` or something similar, 

Would a configuration option not suffice?

> but I don't know how that best fits with the command list, and your still going to end up looking at an empty list most of the time.

The other pickers have a "no matching ... found". Would a message like that be applicable here? What is the behavior with no notifications at the moment?

tris203:
> > No, because we just grab the unread items list, and marking it read fires the API even and optimistically removes it from the selector without the need to refresh the whole thing.
> 
> Seems like the same logic for the case where the list is already filtered
> 
> > We could change it to a `notification list` and a `notification list all` or something similar,
> 
> Would a configuration option not suffice?
Sure
> > but I don't know how that best fits with the command list, and your still going to end up looking at an empty list most of the time.
> 
> The other pickers have a "no matching ... found". Would a message like that be applicable here? What is the behavior with no notifications at the moment?
an empty list

Seeing as this is fully new functionality, IMO i think it makes sense to get a base design in and iterate on it later once it is being used.
At the moment, the state of this seems to work for my functionality of opening neovim and looking what needs to be done.
Or being working on something, seeing the email notification come in and opening the list in neovim without having to think about switching to the right project first

lua/octo/pickers/telescope/entry_maker.lua
```diff
@@ -694,6 +704,64 @@ function M.gen_from_octo_actions()
...
+      { "#" .. (entry.obj.subject.url:match "/(%d+)$" or "NA") },
+      { string.sub(entry.obj.repository.full_name, 1, 50), "TelescopeResultsNumber" },
+      { string.sub(entry.obj.subject.title, 1, 100) },
+    }
+
+    local displayer = entry_display.create {
+      separator = " ",
+      items = {
+        { width = 2 },
+        { width = 12 },
```
wd60622, line 724: Let's make this smaller.  For instance, search has a max value of 6

https://github.com/wd60622/octo.nvim/blob/b4f00b00dec14fd396be979b04453336d6e83182/lua/octo/pickers/telescope/provider.lua#L541-L546

wd60622:
I think a rebase and address my comment would make this good to go

lua/octo/pickers/telescope/entry_maker.lua
wd60622: An alternative is to make this dynamic sized like with the search as well. The width would be determined then passed to the entry_maker.

wd60622 (approved):
Just one final suggestion on the spacing being too big

wd60622:
Could @pwntester take another look. Your requested change is blocking a potential merge

tris203:
I think your rebase has broken the icons too, but I will have a look

PR #708, opened 2024-12-10
Use plenary sync job to get local file content

### LINKED ISSUE #608: Empty content in diff buffer when trying to fetch from commit at random times
## Issue Description

Type: *bug report*

Sometimes, when reviewing a PR, the left diff is empty when it should not be.
I use local fs so right dif always comes from disk directly, but I suppose this could also happen for right diff.

This does not happen everytime, so I suspect some sort of race condition to be the root cause.

The left diff buffer should show the content of the file at the base commit

1. (Not sure if mandatory) Use local fs
2. Resume a PR review
3. Sometimes, one of the files will be empty when it should not, not all the time, not always the same file, does not seem scoped to a single PR.

### Tell us your environment
Nixos, nvim from unstable
```
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713773202
Run "nvim -V1 -v" for more info
```

I dug a bit on this. I could not go further yet since I am quite unfamiliar with plenary and I could not find documentation.
`utils.get_file_at_commit` sometimes returns an empty array despite commit being present locally.

When I print output and stderr [here](https://github.com/pwntester/octo.nvim/blob/955f0fdd03b0da23ee1c3368d80bf806f0e36fa5/lua/octo/utils.lua#L247-L251), both tables are empty.

I also printed error code and signal (the second and third arg of the `on_exit` cb), `0` for both which seem normal.

Somehow, the output of `git show rev:file` from Plenary remains empty, which I have no idea why.

I will update the issue if I find anything else, right now the only workaround I found was to close the review buffers/restart nvim and re open the PR.

GuillaumeLagrange:
Fixes #608

I gave up on trying to troubleshoot async plenary, and switched to sync, this fixes the empty diff buffer for local mode that is quite hard to reproduce.
Since this is a local git command, and, impact should be minimal, and the only caller of this function waits for the job to be done anyway by calling vim.wait on a condition that is done by the callback

### DIFF
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -342,13 +342,10 @@ function M.get_file_at_commit(path, commit, cb)
     enable_recording = true,
     command = "git",
     args = { "show", string.format("%s:%s", commit, path) },
-    on_exit = vim.schedule_wrap(function(j_self, _, _)
-      local output = table.concat(j_self:result(), "\n")
-      local stderr = table.concat(j_self:stderr_result(), "\n")
-      cb(vim.split(output, "\n"), vim.split(stderr, "\n"))
-    end),
   }
-  job:start()
+  local result = job:sync()
+  local output = table.concat(result, "\n")
+  cb(vim.split(output, "\n"))
 end
 
 function M.in_pr_repo()


wd60622 (approved):
LGTM

Issue #675, opened 2024-11-10
Make use of telescope multiselect


wd60622:
No description provided.

PR #722, opened 2024-12-15
Support label multiselection

wd60622:

Allows user to allow multiselect for adding and removing labels

Related to #675

### Describe how you did it

- Adapt the graphql command to pass list of labelIds
- Generalize telescope logic to grab multiple selection and check if non-empty
- Use callback on list of labels including list of single label

### Describe how to verify it

For telescope, use `<Tab>` to select multiple labels (while adding or removing) then press enter to confirm. If Tab is not pressed, then the behavior will be the same as before.

### Special notes for reviews

Telescope only. However, fzf had support before but the cb was changed to support batch label add.

### Checklist


wd60622:
Luckily reviewers and assignees is very similar. Will address in the near future as well

PR #752, opened 2024-12-30
fix: trim off .git from remote url properly

PeterCardenas:
We were attempting to trim off the .git from the remote url to get the
remote repo name with owner. However, when the repo name ended with git,
that part of the repo name was trimmed off, e.g. cmp-git became cmp.
This properly escapes the `.` to only trim off `.git`.


wd60622 (approved):
Thanks @PeterCardenas !

PR #775, opened 2025-01-02
fix: create pr with upstream remote now works

### LINKED ISSUE #400: PR creation doesn't recognize remote branch
## Issue Description

*bug report*

I wanted to create a pull request with octo, but for some reason there was a message displayed, that the current branch has no remote and if I wanted to push the current one. The current brach was pulled from gh and also did pushes right before, so it was clearly connected. After restarting nvim I was able to create the pr.

The pr creation was successful most of the time, so I didn't expect it to fail. Of course restarting nvim isn't that big of a deal, but it's just not convenient.

I had the bug before, but sadly I don't know how to recreate it yet.

### Tell us your environment

Windows 11
gh version 2.29.0
octo latest

Maybe this is an issue of gh cli?

Dronakurl:

I wanted to create a pull request with octo, but for some reason there was a message displayed, that the current branch has no remote and if I wanted to push the current one. 
The reason was that the command pulls remote branches from the remote that takes precedence. 
That is the upstream repo. Therefore, the local branch of course does not exist in that remote.

Fixes #400

### Describe how you did it

I changed the `pr create` command so that it first tries to get from the remote `origin` not according to the precedence.

### Describe how to verify it

This very PR was created with Octo :-)

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -931,7 +931,12 @@ function M.create_pr(is_draft)
     end
     repo = remotes[remote_idx].repo
   else
-    repo = utils.get_remote_name()
+    -- Override the precedence of get_remote, because otherwise upstream is selected
+    -- and the check if the local branch creates on the repo fails.
+    repo = utils.get_remote_name { "origin" }
+    if not repo then
+      repo = utils.get_remote_name()
+    end
     if not repo then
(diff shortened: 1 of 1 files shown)


wd60622 (approved):
Looks good. Let's go for it! Thanks for the PR

Issue #587, opened 2024-09-01
`comment add` doesn't work for review's visual block


wd60622:
The `:Octo comment add` command works in Review diffs. However, the visual block doesn't work and I get "No range allowed" error which I find to be a confusing experience. 

To reproduce: 
1. Go to a PR diff
2. Select a block of code
3. Type `:Octo comment add`

Using the shortcut of `<leader>ca` works fine with a block selection which might be the intended usage.

Issue #303, opened 2022-06-18
[feature] make 'Octo comment add' work for visual


alex-popov-tech:
## Issue Description

Type: *feature request*

### Describe what feature you want
Please make 'Octo comment add' work on visual, it will be more consise and understandable than now using OctoAddReviewComment. Also relevant for suggestions.

### Anything else we need to know?

Related to https://github.com/pwntester/octo.nvim/discussions/301

alex-popov-tech:
its working now like a charm, thank you <3

PR #807, opened 2025-01-17
fix: make comment ranges work

ldelossa:
Previous to this comment highlighting several lines and running the command "Octo comment add" would never result in a multi-line comment.

This is seemingly because as soon as you enter the command line with ':' Neovim is switched back into normal mode and the "v" and "." attributes to `vim.fn.line` no longer register for the line range.

Instead, we can use the explicit start and end marks for a selected range to capture the currently selected range despite being in normal mode at time of command run.

It is now possible to highlight several lines in a review and then call "Octo comment add" to create a multi-line comment. 

### Checklist

### DIFF
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -1570,8 +1570,8 @@ function M.get_lines_from_context(calling_context)
     line_number_start = vim.fn.line "."
     line_number_end = line_number_start
   elseif calling_context == "visual" then
-    line_number_start = vim.fn.line "v"
-    line_number_end = vim.fn.line "."
+    line_number_start = vim.fn.line "'<"
+    line_number_end = vim.fn.line "'>"
   elseif calling_context == "motion" then
     line_number_start = vim.fn.getpos("'[")[2]
     line_number_end = vim.fn.getpos("']")[2]
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -33,7 +33,7 @@ end
 function M.setup()
   vim.api.nvim_create_user_command("Octo", function(opts)
     require("octo.commands").octo(unpack(opts.fargs))
-  end, { complete = require("octo.completion").octo_command_complete, nargs = "*" })
+  end, { complete = require("octo.completion").octo_command_complete, nargs = "*", range = true })
   local conf = config.values
 
   local card_commands


wd60622:
Were there any issues for this?

ldelossa:
Only relevant issue is: https://github.com/pwntester/octo.nvim/issues/303

But there seems to be a lot of refactoring since then. Possible it broke or most people are using keymaps to create multiline comments? I'm not sure. 

I'd want the command to work for multi-line comments, as I already have far too many keymaps and just typing things out are becoming an easier approach.

wd60622:
Awesome. Let me take a look tomorrow

lua/octo/utils.lua
```diff
@@ -1570,8 +1570,8 @@ function M.get_lines_from_context(calling_context)
     line_number_start = vim.fn.line "."
     line_number_end = line_number_start
   elseif calling_context == "visual" then
-    line_number_start = vim.fn.line "v"
-    line_number_end = vim.fn.line "."
+    line_number_start = vim.fn.line "'<"
+    line_number_end = vim.fn.line "'>"
```
wd60622, line 1574: Awesome. Works like a charm!

wd60622 (approved):
Works like a charm. Nice live quality improvement!

lua/octo/utils.lua
wd60622: Closes #587

wd60622:
Closes #587

lua/octo/utils.lua
raulchen, line 1574: hey @wd60622 @ldelossa , looks like this PR broke non-range selection. 
Because the caller always pass "visual" as the calling_context, when not in the visual mode, this function will always return `0, 0`.

ldelossa: When you say "non range selections" what do you mean? Can you give me steps to reproduce and ill take a look

raulchen: I meant in the normal mode. When I add comment without any visual selection, the comment buffer can still pop up. But when I save the comment, it will fail with an error "failed to create thread". 
I dug into this issue and it turned out that the bug is because the line numbers are 0.

raulchen: it's with a keybind

ldelossa: Yup I see. Thanks for info. I think youre onto it with maybe using "visual" argument in every scenario.

ldelossa: Opened a PR with a proposed fix for this: https://github.com/pwntester/octo.nvim/pull/825

PR #814, opened 2025-01-22
`Octo pr checkout` uses picker when not in Octo buffer

wd60622:

When not in a Octo buffer, the command will use the PR picker to select a PR to checkout.

https://github.com/user-attachments/assets/422fa43e-5c25-40d4-ab01-64fbbe54b121

### Describe how you did it

Very easy following #803

### Describe how to verify it

1. Open a non-Octo buffer
2. Run `:Octo pr checkout`

### Special notes for reviews

Only works for telescope

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -205,6 +205,11 @@ function M.setup()
         local bufnr = vim.api.nvim_get_current_buf()
         local buffer = octo_buffers[bufnr]
         if not buffer or not buffer:isPullRequest() then
+          picker.prs {
+            cb = function(selected)
+              utils.checkout_pr(selected.obj.number)
+            end,
+          }
           return
         end
         if not utils.in_pr_repo() then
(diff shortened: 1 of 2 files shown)


ldelossa:
This is dope @wd60622 

I bit of a thought about this. It would be cool if "checkout" means it also loads up the review layout. 

Sometimes, you may want to look at a PR, but have no desire to add a review. You may just want to browse the changes and read comments. Right now, this required you to start a review, right? 

If so, it would be nice to make "checkout" command load up what is currently the review layout, and then make "review.start" actually start a review. In other words, the review.start functionality today just becomes a subset of the mode youre put in when you do "checkout"

wd60622:
Thanks. I was thinking itd be helpful

That sounds useful too. One element is that gh pr checkout is a cli command and its good to stick close to that. For instance, that is still being using here but the pr number is found through the picker. 

Have you taken a look at Octo pr changes? It has some elements of what you mentioned. There is no picker aspect to it at the moment though

wd60622:
I'll merge this and make some issues based on this comment @ldelossa

ldelossa:
Sounds good! this is a solid inclusion :)

PR #820, opened 2025-01-25
fix: use --slurp with --paginate to concat pages into valid json

ldelossa:
Fixes: #819

The original arguments to `gh api` for certain commands are problematic.

For instance:
```
@@ -85,13 +95,14 @@ end
 function PullRequest:get_changed_files(callback)
   local url = string.format("repos/%s/pulls/%d/files", self.repo, self.number)
   gh.run {
-    args = { "api", "--paginate", url, "--jq", "." },
```

In the above arguments the returned results, if exceeds one pages, will be two separate JSON arrays. E.g. [{...}][{...}].

You cannot feed this directly to `vim.fn.jsondecode` since its not actually valid json, its two json arrays in a single string.

Luckily, `gh` includes the `--slurp` argument to the `api` command which works similiarly to `jq`'s slurp command, concating sequential arrays.

We can use this, with one catch.
The results of any API call with `--slurp` is "an array of pages" and thus the data we want is nested one level deep.

We simply write a `merge_pages` function which extracts all items into a 1d array, which is what Octo currently expects.

### Checklist

### DIFF
--- a/lua/octo/model/pull-request.lua
+++ b/lua/octo/model/pull-request.lua
@@ -63,6 +63,16 @@ end
 
 M.PullRequest = PullRequest
 
+local function merge_pages(data)
+  local out = {}
+  for _, page in ipairs(data) do
+    for _, item in ipairs(page) do
+      table.insert(out, item)
+    end
+  end
+  return out
+end
+
 --- Fetch the diff of the PR
 --- @param pr PullRequest
 function PullRequest:get_diff(pr)
@@ -85,13 +95,14 @@ end
 function PullRequest:get_changed_files(callback)
   local url = string.format("repos/%s/pulls/%d/files", self.repo, self.number)
   gh.run {
-    args = { "api", "--paginate", url, "--jq", "." },
+    args = { "api", "--paginate", url, "--slurp" },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.error(stderr)
       elseif output then
         local FileEntry = require("octo.reviews.file-entry").FileEntry
         local results = vim.fn.json_decode(output)
+        results = merge_pages(results)
         local files = {}
         for _, result in ipairs(results) do
           local entry = FileEntry:new {
@@ -118,13 +129,14 @@ end
 function PullRequest:get_commit_changed_files(rev, callback)
   local url = string.format("repos/%s/commits/%s", self.repo, rev.commit)
   gh.run {
-    args = { "api", "--paginate", url, "--jq", "." },
+    args = { "api", "--paginate", url, "--slurp" },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.error(stderr)
       elseif output then
         local FileEntry = require("octo.reviews.file-entry").FileEntry
         local results = vim.fn.json_decode(output)
+        results = merge_pages(results)
         local files = {}
         if results.files then
           for _, result in ipairs(results.files) do


ldelossa:
hmm, this could be seen as a breaking change, but only if you're using a `gh` cli older then when the `--slurp` argument was introduced: https://github.com/cli/cli/pull/8620

I'll look for advice on how to handle that.

wd60622:
Thanks for the PR @ldelossa

Let me take a look at the code. As for the `gh` cli version, `--slurp` is being used in many places. I wouldn't concern about that

lua/octo/model/pull-request.lua
```diff
@@ -118,13 +129,14 @@ end
...
   gh.run {
-    args = { "api", "--paginate", url, "--jq", "." },
+    args = { "api", "--paginate", url, "--slurp" },
     cb = function(output, stderr)
       if stderr and not utils.is_blank(stderr) then
         utils.error(stderr)
       elseif output then
         local FileEntry = require("octo.reviews.file-entry").FileEntry
         local results = vim.fn.json_decode(output)
+        results = merge_pages(results)
```
wd60622, line 139: Do you know if this new function can be replaced with this current one? 

https://github.com/pwntester/octo.nvim/blob/5579f207e50cce726a2a5a460f48093ee7558fdb/lua/octo/utils.lua?plain=1#L782-L796

wd60622 (commented):
Wondering if this function already exists

lua/octo/model/pull-request.lua
ldelossa: If im not mistaken those are used for "graphaql" queries and the string returned from  one. This is not the same as getting pages JSON from "gh api" (used without the graphql flag).

ldelossa: Oh no im wrong!
I think youre right. That function can be used (as long as it doesnt require a merge key at all). It looks like the alternative to "--slurp".  

Maybe, to fix the linked issue, we actually just need to use that function for the api queries that are currently broken.

wd60622: Give it a try. Not sure the differences of yours and what is currently there. Just noticed some patterns of use

ldelossa: Will do, stay tuned 📺 

Additionally, if it does work, it would be because "--slurp" isnt being used, it basically splits the raw json string by "\n". Now that "--slurp" exists, maybe its better to move Octo to use this? Just to reduce the need to do the manual string manipulation before handling pages.

ldelossa: Well, I tried using `aggregate_pages` and this does not seem to work. It causes a runtime error:

```
Error executing vim.schedule lua callback: .../pack/deps/opt/octo.nvim/lua/octo/reviews/file-entry.lua:497: Vim:E325: ATTENTION                                                                       
stack traceback:                                                                                                                                                                                      
        [C]: in function 'nvim_buf_set_name'                                                                                                                                                          
        .../pack/deps/opt/octo.nvim/lua/octo/reviews/file-entry.lua:497: in function '_get_null_buffer'                                                                                               
        .../pack/deps/opt/octo.nvim/lua/octo/reviews/file-entry.lua:478: in function 'load_null_buffer'                                                                                               
        .../pack/deps/opt/octo.nvim/lua/octo/reviews/file-entry.lua:473: in function 'load_null_buffers'                                                                                              
        ...site/pack/deps/opt/octo.nvim/lua/octo/reviews/layout.lua:247: in function 'file_safeguard'                                                                                                 
        ...site/pack/deps/opt/octo.nvim/lua/octo/reviews/layout.lua:57: in function 'open'                                                                                                            
        ...m/site/pack/deps/opt/octo.nvim/lua/octo/reviews/init.lua:190: in function 'initiate'                                                                                                       
        ...m/site/pack/deps/opt/octo.nvim/lua/octo/reviews/init.lua:99: in function 'callback'                                                                                                        
        ...m/site/pack/deps/opt/octo.nvim/lua/octo/reviews/init.lua:75: in function 'cb'                                                                                                              
        ...e/nvim/site/pack/deps/opt/octo.nvim/lua/octo/gh/init.lua:163: in function ''                                                                                                               
        vim/_editor.lua: in function <vim/_editor.lua:0> 
```

Something with the construction of the FileEntry. 

Maybe this is why it was not used. I'll need a bit of free time to dig into why this breaks.

wd60622: All good. Thanks for looking into the alternative

ldelossa:
As a quick fix yes. I would like to make aggregate_pages work so we dont have this one off function, but I can do this as a follow up PR if you think thats good.

wd60622:
> As a quick fix yes. I would like to make aggregate_pages work so we dont have this one off function, but I can do this as a follow up PR if you think thats good.

Perfect. I will create an issue for this. There is no rush

wd60622 (approved):
Thanks for the PR @ldelossa !

Issue #817, opened 2025-01-24
[bug] timeout fetching file in PR review window for newly added files


raulchen:
In the PR review UI, when I switch to a newly added file , octo will complains `Error fetching <file_name>"`. 

The following patch fixes the issue for me. I'd be happy to submit a PR if the fix looks good to you. 

```diff
diff --git a/lua/octo/reviews/file-entry.lua b/lua/octo/reviews/file-entry.lua
index 1058602..70f8920 100644
--- a/lua/octo/reviews/file-entry.lua
+++ b/lua/octo/reviews/file-entry.lua
@@ -77,7 +77,9 @@ function FileEntry:new(opt)
     left_binary = opt.left_binary,
     right_binary = opt.right_binary,
     left_lines = {},
+    left_fetched = false,
     right_lines = {},
+    right_fetched = false,
     diffhunks = diffhunks,
     associated_bufs = {},
     viewed_state = pr.files[opt.path],
@@ -205,10 +207,12 @@ function FileEntry:fetch()
   if self.pull_request.local_right then
     utils.get_file_at_commit(right_path, right_sha, function(lines)
       self.right_lines = lines
+      self.right_fetched = true
     end)
   else
     utils.get_file_contents(self.pull_request.repo, right_abbrev, right_path, function(lines)
       self.right_lines = lines
+      self.right_fetched = true
     end)
   end
 
@@ -216,10 +220,12 @@ function FileEntry:fetch()
   if self.pull_request.local_left then
     utils.get_file_at_commit(left_path, left_sha, function(lines)
       self.left_lines = lines
+      self.left_fetched = true
     end)
   else
     utils.get_file_contents(self.pull_request.repo, left_abbrev, left_path, function(lines)
       self.left_lines = lines
+      self.left_fetched = true
     end)
   end
 
@@ -232,7 +238,7 @@ end
 ---Determines whether the file content has been loaded and the file is ready to render
 ---@return boolean
 function FileEntry:is_ready_to_render()
-  return #self.left_lines > 0 and #self.right_lines > 0
+  return self.left_fetched and self.right_fetched
 end
 
 ---Load the buffers.

```

raulchen:
Here we go https://github.com/pwntester/octo.nvim/pull/822

PR #822, opened 2025-01-26
Fix fetch timeout for newly added files

raulchen:

When one side of the file is empty `return #self.left_lines > 0 and #self.right_lines > 0` will always be false. Thus `is_ready_to_render` will never be true. 

Fixes https://github.com/pwntester/octo.nvim/issues/817

### Checklist


ldelossa (approved):
Just outside looking in, fix makes total sense :+1:

GuillaumeLagrange (approved):
lgtm, thanks for the fix

PR #825, opened 2025-01-27
fix: use user command lines ranges

ldelossa:
This PR is in response to: https://github.com/pwntester/octo.nvim/pull/807#discussion_r1929894670

This is a follow to making line selections work via the command line.

The crux of the confusion with this is that entering the command line from any other mode then normal, exits that mode, and enters normal mode.

Octo was trying to determine a selected line range, retroactively, when adding review comments. However, this doesn't really work since if you highlight a line, then want to run the comment add command, you exit visual mode. All 'state' related to visual mode and visual line rages are greedy and never 'reset' to 0 when visual mode is exited.

The "right" way that Vim intends you to figure out if a user command is being ran with a visual selection is by using the options dictionary provided to the command.

Octo has a object/action lookup mechanism which hides this user command options dictionary from the handlers.

This commit tries to fix the above issue without change how every single object/action is called by creating a global variable `OctoLastCmdOpts` which caches the last command's options. This can be used to gather the line range in which teh command was called.

Ideally, the object/action function dispatch function is updated to pass the user-command dictionary to all command handlers, but this is a larger change.

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -11,6 +11,15 @@ local config = require "octo.config"
 local colors = require "octo.ui.colors"
 local vim = vim
 
+-- a global variable where command handlers can access the details of the last
+-- command ran.
+--
+-- this came into existence since some commands like "comment add" need to
+-- understand the line range the comment should be created on.
+-- this is problematic without the command options as you exit visual mode when
+-- enterting the command line.
+OctoLastCmdOpts = nil
+
 local M = {}
 
 local get_current_buffer = function()
@@ -32,7 +41,9 @@ end
 
 function M.setup()
   vim.api.nvim_create_user_command("Octo", function(opts)
+    OctoLastCmdOpts = opts
     require("octo.commands").octo(unpack(opts.fargs))
+    OctoLastCmdOpts = nil
   end, { complete = require("octo.completion").octo_command_complete, nargs = "*", range = true })
   local conf = config.values
 
@@ -536,6 +547,7 @@ function M.octo(object, action, ...)
       utils.error(action and "Incorrect action: " .. action or "No action specified")
       return
     end
+
     res = pcall(a, ...)
     if not res then
       utils.error(action and "Failed action: " .. action)
--- a/lua/octo/reviews/init.lua
+++ b/lua/octo/reviews/init.lua
@@ -329,8 +329,15 @@ function Review:add_comment(isSuggestion)
     return
   end
 
-  -- get visual selected line range
+  -- get visual selected line range, used if coming from a keymap where current
+  -- mode can be evaluated.
   local line1, line2 = utils.get_lines_from_context "visual"
+  -- if we came from the command line the command options will provide line
+  -- range
+  if OctoLastCmdOpts ~= nil then
+    line1 = OctoLastCmdOpts.line1
+    line2 = OctoLastCmdOpts.line2
+  end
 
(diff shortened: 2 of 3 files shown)


ldelossa:
Testing a comment from Octo...

lua/octo/commands.lua
```diff
@@ -32,7 +41,9 @@ end
 
 function M.setup()
   vim.api.nvim_create_user_command("Octo", function(opts)
+    OctoLastCmdOpts = opts
     require("octo.commands").octo(unpack(opts.fargs))
+		OctoLastCmdOpts = nil
```
ldelossa, line 46: There is a catch here. Any async action cannot access OctoLastCmdOpts as it will be reset to nil. This is probably okay, as any async handler which needs it (can't think of any) can make a copy before they launch their async routine.

ldelossa: outdated due to linter, but still stands.

wd60622:
@raulchen can you test this out and see if it fixes

lua/octo/commands.lua
wd60622: I'm not following why a global would be needed. Can't the mode be detected and call appropriate function?

ldelossa: The issue with detecting mode is this: 

- Neovim enters visual mode 
- Lines are selected 
- User issues ":" to get into command line to use "Octo comment add" for example 

Once the user enters the command line Neovim has switched to Normal mode, implicitly. Thats just how Neovim works. Anytime you go to the command line you switch to normal mode. 

So, now in normal mode you run the "Octo comment add" command. The command handler func will NEVER detect youre in visual mode. Because you arent, youre in normal at this point. You can try to detect it with functions like "mode" and "visualmode" but this wont work. The former will report "n" for normal and the latter just reports if visual mode was entered at ANY POINT in the Neovim session. 

The reason there isnt really a way to know if you were in visual mode before entering the command line with ":" is because the arguments to any command handler function ran on the cmd line provide this info via the default arguments. That info is what we are caching in the global variable for the lifetime of the command handler func. You can think of it as the full context of Neovim at the time the cmd handler func is called, regardless of "mode". 

ideally, all cmd handler functions have access to this context, but Octo's dispatch handler discards most of that info and has its own way to dispatch a user command line invocation to its cmd handler funcs. Unfortunately by doing this, we loose key information like "the selected line range at time of command"

wd60622: Interesting. I need to digest this a bit. I have lots of custom commands that work both normal and visual mode

ldelossa: Hmm, I could very well be missing something lol, im good at that :D. 

But as a quick test, you can highlight a few lines in Neovim, then do ": lua print(vim.fn.mode())" and notice youre in Normal mode now.

lua/octo/commands.lua
```diff
@@ -11,6 +11,15 @@ local config = require "octo.config"
 local colors = require "octo.ui.colors"
 local vim = vim
 
+-- a global variable where command handlers can access the details of the last
```
raulchen, line 14: A test comment from Octo

raulchen (approved):
Yes, works fine for me.

lua/octo/commands.lua
wd60622: All good. Lets go with the fix!

wd60622 (approved):

wd60622:
Thanks for the quick fix

Issue #829, opened 2025-01-29
Quick look into PRs


wd60622:
A bit of a thought about this. It would be cool if "checkout" means it also loads up the review layout (but does not start a review just yet)

Sometimes, you may want to look at a PR's details (changed files, comments, comment threads in their relation to source code), but have no desire to add a review. You may just want to browse the changes and read comments. Right now, this required you to start a review, right? 

If so, it would be nice to make "checkout" command load up what is currently the review layout, and then make "review.start" actually start a review. In other words, the review.start functionality today just becomes a subset of the mode youre put in when you do "checkout"

_Originally posted by @ldelossa in https://github.com/pwntester/octo.nvim/issues/814#issuecomment-2620294169_

Issue #823, opened 2025-01-27
Cannot reply to a thread on PR page with `Octo comment add` while a review is in progress


GuillaumeLagrange:
We get an HTTP 422 error, to add a comment we need to first discard or submit the review.

We should be able to add a comment, either immediately or in the review itself.

PR #830, opened 2025-01-29
feat: add octo review browse to browse a pull request

ldelossa:
Add a new command `Octo review browse` which opens the review layout for a PR but does not start a review.

This is handy when you want to look over the changes in a PR but do not necessary want to review it.

When you are in a Review layout initiated by `Octo review browse` you can intuitively convert it into an active review session with `Octo review start`.

Its invalid to call `Octo review browse` after `Octo review start` and the user must either submit or discard the active review to enter the browse state again.

Closes: #829

Provides the ability to browse a PR but not enter a review

A big reason to do this is say, you already submitted a review, but you want to use the (very nice) Review layout to continue looking at other contributor's comments. 

#829

### Describe how you did it
Copied the patterns used for `Octo review start` but instead of submitting graphql mutations to start a review, we simply just get the review threads and initialize the review layout. The abstractions are well done, so this was clean and done without much trouble. 

### Describe how to verify it
Open a PR with `Octo pr edit|list` and issue `Octo review browse` you can then ensure starting a review from the browse state is working with `Octo review start`. 

### Checklist


ldelossa:
going to convert this to draft. 

needs docs update and I want to handle this:

image

A bit better.

phdah:
Tested the draft version out. It's great, but agree about the commenting issue.

wd60622:
Very cool. What is the behavior when you (try to) add a comment while browsing? Does that automatically start a review? Or cause an error?

lua/octo/reviews/init.lua
```diff
@@ -51,6 +51,36 @@ function Review:create(callback)
...
+    args = { "api", "graphql", "-f", string.format("query=%s", query) },
+    cb = function(output, stderr)
+      if stderr and not utils.is_blank(stderr) then
+        utils.error(stderr)
+      elseif output then
+        local resp = vim.fn.json_decode(output)
+        callback(resp)
+      end
+    end,
+  }
```
wd60622, line 73: Interested in using the new pattern?  

```lua
gh.api.graphql {
    query = query,
    opts = {
        cb = cb,
    }, 

}
```

lua/octo/reviews/init.lua
```diff
@@ -51,6 +51,36 @@ function Review:create(callback)
...
 end
 
+-- Get review threads without start a review.
+function Review:populate_threads(callback)
+  local query =
+    graphql("review_threads_query", self.pull_request.owner, self.pull_request.name, self.pull_request.number)
+  -- print query to file /tmp/query
+  local file = io.open("/tmp/query", "w")
+  file:write(query)
+  file:close()
```
wd60622, line 61: Is this just for testing?

lua/octo/reviews/init.lua
ldelossa: Definitely, that is very nice. Will update.

lua/octo/reviews/init.lua
ldelossa: Yup 🤦 

forgot about that. Ill remove.

ldelossa:
@phdah @wd60622 

as far as creating a comment via in "browse" state. 

For now, I think it will just create a notification asking you to start a review. Or, if we want, start a review automatically, but youd need to do a "Octo comment add" once again, since starting a review reloads the layout, and puts you at the first changed file. This maybe not where you wanted to comment. 

As a follow up, id like this to add a non-review commit! Theres some nuance to this however, since it, iirc from when I worked on this, a separate set of APIs. So we need to work in some logic to say "hey is this a non-review commit, if so, use this api and not the review comments api". 

Id like this, but probably in a follow up pr? Thoughts?

GuillaumeLagrange:
I like the idea of not having to start a review in order to post a comment.

I can't think of a good UX in the current octo way of doing things though, i'd like to avoid being prompted everytime we add a comment.

The best idea I have right now is that when a user runs `Octo command add`
- If no review is in progress, add the comment in a standalone way, just like when one clicks here 
image
- If a review is in progress, add the comment to the review

And let the user start a review if they wish to do so using `Octo review`

WDYT ?

While we're touching this part of the code, do you think this can address #823 ?

ldelossa:
@GuillaumeLagrange 

I totally agree. 

Ultimately I think 'Octo command add' in the 'browse' state should add a non-review commit. 
I was originally thinking of implementing this in a second PR since I need to dig into the API changes required for this. 

I was thinking of this PR just providing the initial implementation of the feature, and then following up with API changes to handle "browse state comments". 

Would you rather see it all done in one PR here? I'm not too opposed, would just take a bit longer to get the `Octo pr browse` feature out to users.

GuillaumeLagrange:
I think I am fine with iterating as long as we do not hurt existing UX, which is not the case here.
Feel free to submit the PR without the comment management first!

Btw it's still in draft but let me know if you want a review

phdah:
@ldelossa, would you like to convert the PR from draft to ready? I've been using this for two weeks, and really appreciate the feature. Pushing the non-review sounds good.

ldelossa:
BTW while I am not gone, i do not mind if anyone adds this commit to this PR. Basically, if you try to create a comment, and the review object does not have an "id" (meaning it did not get created yet) just do a vim.notify saying "Start a review to add comments". 

I should be back on this next week 😀

ldelossa:
Alright 

@phdah @wd60622  @GuillaumeLagrange 

This is ready for a review. 
TODO is to implement 'non-review' diff comments when in browse mode. 
I can begin to tackle this, tho as you can see, a bit busy with the 9-5 currently, so expect a bit of slowwwwwness from me.

lua/octo/reviews/init.lua
ldelossa: ah I forgot about this and your other comment. will address this in a bit

lua/octo/commands.lua
```diff
@@ -365,6 +368,14 @@ function M.setup()
...
         local current_review = require("octo.reviews").get_current_review()
         if current_review and utils.in_diff_window() then
+          -- if we have a current_review but no id, we are in browse mode.
+          -- for now, we cannot create comments.
+          -- TODO: implement 'non-review' commits here, which adds a diff commit
+          -- but outside of a review.
+          if current_review.id == -1 then
+            vim.notify("Please start or resume a review first", vim.log.levels.ERROR)
+            return
+          end
```
phdah, line 378: I seem to still get the `gh: Could not resolve to a node with the global id of '-1'` error when trying to do a comment in the `browser` mode. I would assume this should catch that?

phdah: Screenshot 2025-02-21 at 11 25 12

ldelossa: when you do this, do you already have a pending review ?

lua/octo/reviews/init.lua
ldelossa: I updated this bit of code. Nice work on that abstraction, I like it :)

lua/octo/commands.lua
ldelossa: hmmm @phdah I'm having trouble reproducing this.
If you run `octo review discard` and then try again, do you run into this?

phdah: @ldelossa I still get the issue. I can try debugging it on my end.

phdah: Okay, so your update here only works on the `Octo comment add`, but not for the `add_review_comment()`, which is default mapped to `<leader>ca`, found [here](https://github.com/ldelossa/octo.nvim/blob/ldelossa/octo-review-browse/lua/octo/mappings.lua#L133-L135).

I think your id check should be moved to the underlying `Review:add_comment()` method found [here](https://github.com/ldelossa/octo.nvim/blob/ldelossa/octo-review-browse/lua/octo/reviews/init.lua#L347), such that it looks like:
```lua
function Review:add_comment(isSuggestion)
  if self.id == -1 then
    vim.notify("Please start or resume a review first", vim.log.levels.ERROR)
    return
  end
  ...
```
Then the logic is general for all comment adding.

phdah: With that change, it works for using both the `action` and `keybinding` for me at least 😄

ldelossa: Hmm, is this kinda a code smell? The fact we have two different ways to create comments? It appears that `Oct comment add` is context specific and can discover if you're in a review or not? Should we **always** go thru this top-level command and stop allowing calls to `add_review_comment()`? 

@wd60622 @GuillaumeLagrange  would love some feedback on that.

ldelossa: I think handled the scenarios where this error still may occur, going to resolve for now.

ldelossa:
@GuillaumeLagrange @wd60622 

This one is good for a final review

wd60622:
> @GuillaumeLagrange @wd60622
> 
> This one is good for a final review

Will take a look when I can!

phdah (approved):
This looks good to me. Great job on it 😃

wd60622:
Overall looks good and was working fine. I just had the file pane go to the left side when opening. Have you experienced that?

https://github.com/user-attachments/assets/1315ad96-ebed-4d3e-a296-78ea864b2459

ldelossa:
@wd60622 that is odd! I haven't had this happen.  It does this every time? And only for the browse command? It would be really strange since this PR does touch anything about the layout of the review mode.

wd60622:
Yeah, lol. Not sure but it is happening on both reviews and browse. If it is just happening for me, that is not a big deal. I can fix with `<localleader>b` then `<localleader>b` again. Feel free to merge!

wd60622 (approved):

ldelossa:
@wd60622 thank you, I cannot merge :) you'll have to click the button.

wd60622:
Done!

PR #832, opened 2025-01-30
feat: improve visibility of selected entry in review file panel

GuillaumeLagrange:
- Remove nocursorline because I find it very hard to spot my cursor in the review file panel, and add a more visible highlight to the selected entry
- Add hl group for selected entry in file panel
- Update readme list of hl groups
- Make sure we dont override user defined groups

### Before
image

### After
image

I took inspiration from diffview, where it looks like this:
image

[Diffview](https://github.com/sindrets/diffview.nvim/blob/4516612fe98ff56ae0415a259ff6361a89419b0a/lua/diffview/hl.lua#L444) uses the `Type` native (?) hl group and it looked nice on my config, so I went with it.
I'm a bit of a dummy in terms of hl groups, let me know if we can do this better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,43 +476,76 @@ Octo provides a built-in omnifunc completion for issues, PRs and users that you 
 
 ## 🎨 Colors
 
-| Highlight Group             | Defaults to     |
-| --------------------------- | --------------- |
-| _OctoDirty_                 | ErrorMsg        |
-| _OctoIssueTitle_            | PreProc         |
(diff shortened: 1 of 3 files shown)


GuillaumeLagrange:
I think a cleaner way to do this is to add a new category to the table referenced [here](https://github.com/pwntester/octo.nvim?tab=readme-ov-file#-colors) and to make it default to `Type` instead of hardcoding `Type` like a savage :nerd_face:

Will do so

PR #837, opened 2025-02-01
fix: Escape hyphen in ssh_aliases and check for nil in author.user 

### LINKED ISSUE #836: Config: ssh_aliases not capture correctly hostname with hyphen
When using the ssh_aliases configuration option, the hyphen character is not handled correctly. This causes the ssh_aliases to be ignored and the default ssh_aliases to be used instead. This issue is caused by the way the ssh_aliases configuration option is parsed. The hyphen character is not escaped properly, which causes the parser to interpret it as a command line option instead of a value. This issue can be fixed by escaping the hyphen character properly when parsing the ssh_aliases configuration option. 

If i have this config:

```lua
require 'octo'.setup {
  ssh_aliases = { ["github.com-work"] = "github.com" },
}
```

When i execute Octo ccommand, the ssh_aliases are not used and the default ssh_aliases are used instead. This throw an error:

hyphen issue

I will send a PR to fix this issue.

### LINKED ISSUE #838: Error accessing to author and user in PR creation
When creating a PR, the author and user are not set correctly. If both are nil, the command throws an error, and the buffer does not work correctly. Additionally, if the PR buffer is saved, it is not sync with the remote PR.

This is the error message:

Image

While debugging, I noticed that the item object has nil for the user but not for the author. However, when trying to access the user inside the author, it is also nil.

Image

I will send a PR to fix this issue.

richardhapb:

This change fixes the issue where the hyphen in the ssh_aliases replacement is not escaped. This causes the replacement to fail when the alias contains a hyphen.

The second change is related to issue #838, where the check for nil values are not working as expected. I have added a check for nil to author.user.

Fixes #836
Fixes #838

### Describe how you did it

In parse_remote_url function, I have added a `%` before the hyphen in the ssh_aliases replacement.

In write_commit_event function, I have added a check for nil to author.user.

### Describe how to verify it

Test with an ssh alias that contains a hyphen, for example github.com-work
Test with a nil value for author.user

### Special notes for reviews

Little changes

### Checklist

- [X] Passing tests and linting standards
- [X] Documentation updates in README.md and doc/octo.txt

### DIFF
--- a/lua/octo/ui/writers.lua
+++ b/lua/octo/ui/writers.lua
@@ -1347,7 +1347,7 @@ function M.write_commit_event(bufnr, item)
       item.commit.committer.user.login,
       item.commit.committer.user.login == vim.g.octo_viewer and "OctoUserViewer" or "OctoUser",
     })
-  elseif item.commit.author ~= vim.NIL then
+  elseif item.commit.author ~= vim.NIL and item.commit.author.user ~= vim.NIL then
     table.insert(vt, {
       item.commit.author.user.login,
       item.commit.author.user.login == vim.g.octo_viewer and "OctoUserViewer" or "OctoUser",
--- a/lua/octo/utils.lua
+++ b/lua/octo/utils.lua
@@ -221,6 +221,7 @@ function M.parse_remote_url(url, aliases)
   end
 
   for alias, rhost in pairs(aliases) do
+    alias = alias:gsub("%-", "%%-")
     host = host:gsub("^" .. alias .. "$", rhost, 1)
   end
   if not M.is_blank(host) and not M.is_blank(repo) then
--- a/lua/tests/plenary/utils_spec.lua
+++ b/lua/tests/plenary/utils_spec.lua
@@ -67,3 +67,55 @@ describe("Utils escape_char(): ", function()
     eq(expected, this.escape_char(input))
   end)
 end)
+
+describe("Utils parse_remote_url(): ", function()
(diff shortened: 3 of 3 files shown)


wd60622:
Hi @richardhapb 

Thanks for the PR?
Are these from ghost users / deleted accounts? Do you have a PR to share with to check?

GuillaumeLagrange (approved):
Lgtm! Thanks for the PR

richardhapb:
@wd60622

Actually, I’m looking at the trace of the error. I haven’t found the exact moment when the values take a nil value yet, but I think it happens in the response from the API. Maybe it’s because I log in to GitHub with a token? It ocurres in my two accounts.

Note: I added some tests for the parse_remote_url function.

Thanks for the plugin!

richardhapb:
@wd60622

Actually, I’m looking at the trace of the error. I haven’t found the exact moment when the values take a nil value yet, but I think it happens in the response from the API. Maybe it’s because I log in to GitHub with a token? It ocurres in my two accounts.

Note: I added some tests for the parse_remote_url function.

Thanks for the plugin!

GuillaumeLagrange:
@wd60622 do we merge this or do you want to dig a bit deeper into the root cause of the nil user ?
@richardhapb if you have an openly accessible PR where we can reproduce the issue it would help :hugs:

wd60622:
Let's merge and iterate!

wd60622 (approved):
Thanks for the PR @richardhapb

PR #847, opened 2025-02-04
Include mode of command in config and make submit window commands work in insert mode

wd60622:

Allow uses to optionally pass the mode of the command in the config

### Describe how to verify it

See that all of the previous commands are working as before and that the submit window commands work in insert mode as well.

### Checklist


lua/octo/config.lua
```diff
@@ -271,16 +271,16 @@ function M.get_default_values()
...
+        request_changes = { lhs = "<C-r>", desc = "request changes review", mode = { "n", "i" } },
+        close_review_tab = { lhs = "<C-c>", desc = "close review tab", mode = { "n", "i" } },
       },
       review_diff = {
         submit_review = { lhs = "<localleader>vs", desc = "submit review" },
         discard_review = { lhs = "<localleader>vd", desc = "discard review" },
-        add_review_comment = { lhs = "<localleader>ca", desc = "add a new review comment" },
-        add_review_suggestion = { lhs = "<localleader>sa", desc = "add a new review suggestion" },
+        add_review_comment = { lhs = "<localleader>ca", desc = "add a new review comment", mode = { "n", "x" } },
+        add_review_suggestion = { lhs = "<localleader>sa", desc = "add a new review suggestion", mode = { "n", "x" } },
```
GuillaumeLagrange, line 283: We should update default mappings here: https://github.com/pwntester/octo.nvim/blob/17bdb91ed22f4b6378f88c1791d0b85a37142d1a/README.md?plain=1#L171-L303

lua/octo/utils.lua
```diff
@@ -1658,11 +1658,8 @@ function M.apply_mappings(kind, bufnr)
         value.desc = ""
       end
       local mapping_opts = { silent = true, noremap = true, buffer = bufnr, desc = value.desc }
-      vim.keymap.set("n", value.lhs, mappings[action], mapping_opts)
-      -- TODO: These should probably be part of the config
-      if action == "add_review_comment" or action == "add_review_suggestion" then
-        vim.keymap.set("x", value.lhs, mappings[action], mapping_opts)
-      end
+      local mode = value.mode or "n"
+      vim.keymap.set(mode, value.lhs, mappings[action], mapping_opts)
```
GuillaumeLagrange, line 1662: We should mark this as a breaking change using `!` in the commit message since users (like me 🤓) who manually set `add_review_comment/suggestion` mappings in their config will no longer have their mapping work in visual mode without them explicitely adding the mode.

It's much better this way, but still confused me

GuillaumeLagrange (approved):
Olgtm

lua/octo/utils.lua
wd60622: What is the format of this? I will do upon squash

lua/octo/config.lua
wd60622: done

PR #850, opened 2025-02-06
Better types and minor refactor/cleanup

TheLeoP:

I want to contribute more to the codebase, but it's a bit daunting without knowing it. Type annotations make it easier and more manageable. But, A lot of the types come from the Github API, so I also added a script to generate them from the Graphlq schema definition (and a the generated file). I didn't know Github also had a REST API before this PR, so those types may need to wait for now :p

NONE

### Describe how you did it

Used the Grahpql specificacion to generate LuaCAT annotations for the Github Grapgql schema. I used those types to better type (and caught a couple of error while doing it) some parts of the codebase

### Describe how to verify it

I think I shouldn't have change the behavior of the plugin, but something may have slipped through my refactors. I guess the test should catch it if it happened

### Special notes for reviews

Ignore the huge line count, it's mostly because of the huge generate file with the type definitions.

The script on `scripts/generate_gh_types.lua` needs to be run from within Neovim by doing `:source` with that file opened. It'll generate the file `./lua/octo/_meta/gh.lua`

### Checklist


lua/octo/_meta/gh.lua
wd60622, line 1: I am not familar with this file. Is this generated by the new script? How do we ensure that it is up to date?

wd60622: Is this worth a 60% increase in size of the plugin?

TheLeoP: > I am not familar with this file. Is this generated by the new script?

Yes

> How do we ensure that it is up to date?

It'll only need to be regenerated if good has a breaking change in their graphql API. In which case, the plugin would need to be updated anyway. 

> Is this worth a 60% increase in size of the plugin?

It's not my decision to make, but, in my opinion, it is. The file only contains comments (it doesn't contain any kind of runtime functionality) and doesn't need to be manually maintained, only generated using a script. 

In exchange, it'll allow to avoid type checking problems in the future (like not checking for vim.NIL when needed)

wd60622: Cool. So is it for the users of the API? I think that would mainly be the developers. 
Can you show a clip showing what it does

TheLeoP: I don't understand your question. Could you elaborate?

wd60622: What happens now that this is implemented? Who will see that change?

TheLeoP: Any developer of octo.nvim can use the type annotations generated and added by this PR. Lua-language-server will give warnings and autocompletion about them and it'll help to add types to the functions signatures in all of the plugin. This will avoid common issues like forgetting to check if a value is `vim.NIL` before indexing it, because lua-language-server will warn you about it

MuntasirSZN (approved):

MuntasirSZN:
What happened @TheLeoP ?

TheLeoP:
I hadn't received any more feedback in a while and the plugin went through some major refractors, so I opted for closing the PR

PR #851, opened 2025-02-06
Improve buffer behavior after thread creation

GuillaumeLagrange:
Fix error `gh: Could not resolve to a node with the global id of '-1'` when creating a comment from a thread buffer that was just used to create a new thread without closing it first

### Describe how you did it
- Improved review typing
- Improved behavior of `Octo comment add` by ensuring the thread buffer was visible, and therefore that existing comment had been loaded, BEFORE trying to create a new comment
- Fixed the error of creating a new comment in a fresh thread, which was caused by the octo-buffer thread metadata not being updated with ids of created thread after the gh api

### Describe how to verify it
- Open a review
- On a line without a comment, run `Octo comment add`
- Save a new comment
- Re-run `Octo comment add` to add a reply to the comment you just created
- See the comment being registered without error, when the error `gh: Could not resolve to a node with the global id of '-1'` would previously be shown

### Checklist

### DIFF
--- a/lua/octo/autocmds.lua
+++ b/lua/octo/autocmds.lua
@@ -55,7 +55,7 @@ function M.setup()
       group = "octo_autocmds",
       pattern = { "*" },
       callback = function()
-        require("octo.reviews.thread-panel").show_review_threads()
+        require("octo.reviews.thread-panel").show_review_threads(false)
       end,
     })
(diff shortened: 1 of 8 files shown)


wd60622 (approved):
Good stuff

PR #860, opened 2025-02-16
fix(fzf): upstream change "" to nil when no selected

phanen:
`entry_str` don't need to be checked as empty string when nothing selected, since it's send as `nil` and avoided by:
https://github.com/ibhagwan/fzf-lua/blob/f8ccd73a7a3a0c9adae935441f85e83d5a989efd/lua/fzf-lua/previewer/builtin.lua#L355

Related: https://github.com/ibhagwan/fzf-lua/pull/1829

### Checklist


wd60622:
Thanks for the PR @phanen

PR #862, opened 2025-02-16
feat(snacks): async notifications

tris203:
Makes the notifications picker use the async system properly

This might need some iteration, the snacks picker doesn't really work with json very well

### DIFF
--- a/lua/octo/gh/init.lua
+++ b/lua/octo/gh/init.lua
@@ -148,6 +148,9 @@ function M.run(opts)
       table.insert(opts.args, header)
     end
   end
+  if opts.return_args_only then
+    return opts.args
+  end
   local job = Job:new {
     enable_recording = true,
     command = config.values.gh_cmd,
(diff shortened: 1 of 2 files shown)


lua/octo/gh/init.lua
```diff
@@ -148,6 +148,9 @@ function M.run(opts)
       table.insert(opts.args, header)
     end
   end
+  if opts.return_args_only then
+    return opts.args
+  end
```
wd60622, line 153: Maybe the function should be broken into two. 
- Creating the args 
- running the command

gh.run with `return_args_only` seems confusing

wd60622: Plus this would making debugging / testing a lot easier if we can just swap out
`gh.run` with `gh.create_args` but have the same signature

williambdean: Hi @tris203, are you still interested in this PR?

tris203: Yes I think it's necessary to move the snackers picker along

However, iirc when I looked the JSON/object parsing was limited to make the best benefit of it

Let me relook at where I got

PR #884, opened 2025-02-23
allow for gh run commands as well as previous run function

wd60622:
The function `gh.run` didn't allow for using the subcommands of GitHub CLI's `run` (Dynamic subcommands from #810)

This change allows for both. For example, `gh.run { args = { "pr", "list" } }` is backwards compatible. 

But `gh.run.list {}`, `gh.run.view { number }`, etc is now also possible.

### DIFF
--- a/lua/octo/gh/init.lua
+++ b/lua/octo/gh/init.lua
@@ -347,6 +347,7 @@ local create_subcommand = function(command)
 
   setmetatable(subcommand, {
     __call = function(_, opts)
+      --- Allow for backwards compatibility with the old API gh.run { ... }
       if command == "run" then
         return run(opts)
       end
--- a/lua/octo/gh/init.lua
+++ b/lua/octo/gh/init.lua
@@ -141,7 +141,7 @@ end
 ---Run a gh command
 ---@param opts RunOpts
 ---@return string[]|nil
-function M.run(opts)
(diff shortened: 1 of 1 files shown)


wd60622:
CC: @GustavEikaas

GustavEikaas:
I think I understand what you are trying to do but im not entirely sure what i'm doing wrong. The following code does not work it seems

```lua
  print(vim.inspect(gh.run.list)) -> function 1
  gh.run.list {
    args = { "--json", run_list_fields, "-L", workflow_limit },
    cb = function(output, stderr)
    end
}

```

wd60622:
> I think I understand what you are trying to do but im not entirely sure what i'm doing wrong. The following code does not work it seems
> 
> 
> 
> ```lua
> 
>   print(vim.inspect(gh.run.list)) -> function 1
> 
>   gh.run.list {
> 
>     args = { "--json", run_list_fields, "-L", workflow_limit },
> 
>     cb = function(output, stderr)
> 
>     end
> 
> }
> 
> 
> 
> ```

Close. Any args or parameters github CLI are just passed as table. There is restricted key "opts" which passes to (previous) gh.run

Try 
```lua
gh.run.list {
  json = fields,
  L = limit,
  opts = { cb = cb },
}
```

PR #890, opened 2025-02-23
refactor: gh run command

GustavEikaas:
Refactored `gh.run` to `gh.run.list`

### DIFF
--- a/lua/octo/workflow_runs.lua
+++ b/lua/octo/workflow_runs.lua
@@ -646,43 +646,42 @@ local workflow_limit = 100
 
 local run_list_fields = "conclusion,displayTitle,event,headBranch,name,number,status,updatedAt,databaseId"
 
-local function get_workflow_runs_sync(co)
+local function get_workflow_runs_sync()
   local lines = {}
-  gh.run {
-    args = { "run", "list", "--json", run_list_fields, "-L", workflow_limit },
-    cb = function(output, stderr)
-      if stderr and not utils.is_blank(stderr) then
-        vim.api.nvim_err_writeln(stderr)
-        utils.error "Failed to get workflow runs"
-      elseif output then
-        local json = vim.fn.json_decode(output)
-        for _, value in ipairs(json) do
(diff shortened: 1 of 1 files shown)


GustavEikaas:
@wd60622 I understood what you meant now. I dont think gh.run.view <id> would be feasible to replace without making changes to gh.run handler function. Did some dabbling but it always ended with `gh run view -- <id>`. Could optionally change the `create flag` function to ignore the argument if the `#key == 0`

wd60622:
You can just do gh.run.view { id }. No need to have a blank key. Lua does that internally

wd60622:
I can write up an example if you want

wd60622 (commented):
What does changing to sync do?

GustavEikaas:
Just lets me remove the `coroutine.yield` and `coroutine.resume` that I had previously

lua/octo/workflow_runs.lua
```diff
@@ -307,10 +305,11 @@ local function get_logs(id)
     end
 
     local sanitized_name = node.id:gsub("/", ""):gsub(":", ""):gsub(">", "")
-    --Make more than 3 consecutive dots at the end of line into ...
+    --Make more than 3 consecutive dots at the end of line into */. This avoids a bug with unreliable filename endings
     local sanitized_job_id = node.job_id:gsub("/", ""):gsub(":", ""):gsub("%.+$", "*/")
     local file_name = string.format("%s_%s.txt", node.number, sanitized_name)
-    local path = sanitized_job_id .. file_name
+    local path = vim.fs.joinpath(sanitized_job_id, file_name)
```
GustavEikaas, line 311: This fixes a bug where it would fail to extract the correct file if there were no `/` at the end of the directory name

GustavEikaas:
> I can write up an example if you want

I see now, numerical keys doesnt interfere with the flags magic. Both commands are now refactored and working

wd60622 (approved):
Cool. Is this good to go?

GustavEikaas:
Yep! LGTM🚀

PR #911, opened 2025-03-02
Add `:Octo pr runs` command

### LINKED ISSUE #910: Get workflow runs for a PR
Should be able to add the branch flag to the current function call here: 

https://github.com/pwntester/octo.nvim/blob/03052ea89ffa9f4c99c0d3f261e453a719160075/lua/octo/workflow_runs.lua?plain=1#L652-L656

then expose it here: 

https://github.com/pwntester/octo.nvim/blob/03052ea89ffa9f4c99c0d3f261e453a719160075/lua/octo/workflow_runs.lua?plain=1#L703-L705

as well as in the commands.lua

Though it would be using the workflow logic, I think `Octo pr runs` might make sense as a command.

Thoughts on this? @GustavEikaas

wd60622:

Expose the `branch` in the `list` function which is used in new command `:Octo pr runs` to get the runs associated with the PR.

In this case, the value in the picker are all the same. Would it make sense to show the workflow name instead of the branch name? That is what comes from `gh run list --branch pr-runs --json workflowName`

For instance, [this run](https://github.com/pwntester/octo.nvim/actions/runs/13613561550) would be labeled "Sync Closing Labels" 

CC: @GustavEikaas

Closes #910 

### Checklist


GustavEikaas (commented):
Looks good!

One thing however;

You are passing branch to the filter for displaying workflows related to a pr. This will return all workflows run on this branch. Including the ones that have been run with triggers like

workflow_dispatch:

Is that intentional?

EDIT:
It makes sense IMO just sanity checking.

GustavEikaas (approved):

GustavEikaas:
> In this case, the value in the picker are all the same. Would it make sense to show the workflow name instead of the branch name? That is what comes from gh run list --branch pr-runs --json workflowName
For instance, [this run](https://github.com/pwntester/octo.nvim/actions/runs/13613561550) would be labeled "Sync Closing Labels"

Yes definitely, very confusing when you list wf runs for a PR and they are all named the same

wd60622:
I was playing around with it and came to same conclusion in 9aec9

wd60622:
> You are passing branch to the filter for displaying workflows related to a pr. This will return all workflows run on this branch. Including the ones that have been run with triggers like
> 
> workflow_dispatch:
> 
> Is that intentional?


Yeah, I think that is probably fine

wd60622:
I think it might make sense to have name even for the larger `run list`? What are your thoughts? Maybe prepend `name` to what is already being displayed? Bit separate so can address that some other time..

GustavEikaas:
What we could do as an overall improvement is this
```lua
 local display
      if opts.branch == nil then
        display = string.format("%s (%s), value.name, value.headBranch)
      else
        display = value.name
      end
```

GustavEikaas:
Awesome! LGTM🚀

wd60622:
I've added that in in 483e424 . I tend to search by the "name" on the workflow

Let me know how this looks to you

wd60622:
Thanks for the feedback @GustavEikaas !

GustavEikaas:
Remember to update readme

PR #916, opened 2025-03-04
refactor!: update deprecated highlight VertSplit to WinSeparator

disrupted:

`hl-VertSplit` is deprecated in Neovim core. Help advises to use `hl-WinSeparator` instead. This is a breaking change because of the rename.

NONE

### Checklist


wd60622 (approved):
Good catch. Thanks!

PR #927, opened 2025-03-10
refactor: switch remaining `vim.fn.json_...` calls to `vim.json`

disrupted:

I noticed that a few `vim.fn.json_...` calls remain in the codebase while most of them already use the newer `vim.json` API. Not sure if something else needs to be done here.

NONE

### Checklist


wd60622:
Thanks @disrupted !

Issue #854, opened 2025-02-07
[bug] keymaps lost when switching to a different tab from the review-diff view and switch back


raulchen:
Reproduce:
1. `Oct <pr_url>` to open a PR
2. start a diff review, which will start a new tab.
3. in the review-diff view, I can use Oct keymaps, e.g. "<localleader>ca" to add a comment.
4. When I switch to the previous tab and switch back, the keymaps are lost.

raulchen:
Dug into the code, it looks like it's this [_detach_buffer](https://github.com/pwntester/octo.nvim/blob/940d7af617015fd5c65f707a267d3daa29be480e/lua/octo/reviews/file-entry.lua#L537) function that delete the keymaps when switching away from the review-diff view. 

I don't really understand why it's necessary to delete the kaymaps. IIUC, it's only needed when `use_local_fs=true`?

wd60622:
Any idea @GuillaumeLagrange ?

GuillaumeLagrange:
I've observed this indeed, I'll try and submit a PR over the weekend if nobody does before me

raulchen:
hey @williambdean @GuillaumeLagrange, I submitted a fix PR https://github.com/pwntester/octo.nvim/pull/1050. Could you please take a look? thanks.

PR #1050, opened 2025-04-08
Fix lost keymaps when switching tabs

raulchen:

* When reentering the octo tab, re-apply keymaps. 
* When `use_local_fs`, there is no need to remove keymaps in the first place. 

Fixes https://github.com/pwntester/octo.nvim/issues/854

### Describe how to verify it

Manually tested. 

### Checklist


lua/octo/reviews/layout.lua
```diff
@@ -256,14 +256,14 @@ function Layout:on_enter()
   end
 
   local file = self:get_current_file()
-  if file then
+  if file and config.values.use_local_fs then
```
williambdean, line 259: I have some issues with my local file system. but this change is only for local_fs?

lua/octo/autocmds.lua
```diff
@@ -66,6 +66,13 @@ function M.setup()
       require("octo.reviews").close(tonumber(vim.fn.expand "<afile>"))
     end,
   })
+  define({ "TabEnter" }, {
+    group = "octo_autocmds",
+    pattern = { "*" },
+    callback = function()
+      require("octo.reviews").on_tab_enter()
```
williambdean, line 73: What triggers this the first time?

williambdean (commented):
Thanks for the PR!

lua/octo/reviews/layout.lua
raulchen: this change is not needed to fix the bug. because `on_tab_enter` will restore the keymaps. 
The reason I still added this change is because when not using local fs (i.e. octo uses its own buffers), there is no need to delete the keymaps in the first place.

raulchen:
> Thanks for the PR! Overall, think this looks good. I tried the issue summary to reproduce the problem.
> 
> I started review, switched tab, then switched back to the octo tab. The keymaps were not lost.
> 
> Is that a success? I don't used the `local_fs` option, so I don't know if I am correctly testing that part.

yeah, your verification is correct. you don't need to enable local_fs to verify the fix.

raulchen:
@williambdean hey, is the PR good to go?

williambdean (approved):
Thanks for the bug fix!

PR #1065, opened 2025-04-26
feat: add command for marking notification as done

### LINKED ISSUE #1064: Add ability to mark a notification as done
### What feature you want

I want to be able to mark notifications as done, rather than read. This removes the notification from the GitHub UI, which helps to keep notifications tidy. 

In my workflow I use this functionality to reduce the amount of notifications which I'm looking at. 

Do you have a work around?

Workaround is to mark notifications as read, and then use the GitHub UI to mark all stale notifications as done in bulk. 

Does this build upon an existing feature?

It builds upon the telescope picker, should be a simple extension of the existing code for marking notifications as done.

Sharn-konet:

These changes add a new binding `<localleader>nd` inside of the notification picker which marks a notification as done. 

It's needed because this action clears the notification from the GitHub UI, the current binding for reading a notification leaves the notification in the UI, even though it removes the notification from octo. 

Fixes #1064 

I created that issue since the CONTRIBUTING.md file mentioned that each PR should be linked to an issue. Let me know if that was not the intention. 

### Describe how you did it

I've added a new function which sends a delete request on the picked notification, rather than a PATCH request marking it as read. From the [GitHub API docs](https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-done) this is how you mark a notification as complete. Doing so removes that notification from the GitHub UI. 

### Describe how to verify it

1. Have a notification available in the `Octo notification list` command. 
2. Verify that the notification is available also at [the notification tab](https://github.com/notifications)
3. In the `Octo notification list` command, use `<localleader>nd`
4. Observe the notifications removal from the picker.
5. Look at the GitHub UI and verify that the notification no longer exists.

Optional: you can also try reading a notification from `Octo notification list` first to verify the problem, which is that the existing command does not remove the notification from the notification tab.

### Special notes for reviews

While implementing this I had an idea that it may be better to replicate the behaviour for reading notifications in the GitHub UI. So, rather than reading notifications removing them from the picker as it does currently, it changes the colour of the icon to show that it's "read". Let me know if that sounds better and I can raise a separate issue. 

### Checklist

### DIFF
--- a/lua/octo/pickers/telescope/provider.lua
+++ b/lua/octo/pickers/telescope/provider.lua
@@ -1330,6 +1330,25 @@ local function mark_notification_read()
   end
 end
 
+local function mark_notification_done()
+  return function(prompt_bufnr)
+    local current_picker = action_state.get_current_picker(prompt_bufnr)
+    current_picker:delete_selection(function(selection)
+      local url = string.format("/notifications/threads/%s", selection.thread_id)
+      gh.run {
+        args = { "api", "--method", "DELETE", url },
+        headers = { "Accept: application/vnd.github.v3.diff" },
+        cb = function(_, stderr)
+          if stderr and not utils.is_blank(stderr) then
+            utils.error(stderr)
+            return
+          end
+        end,
+      }
+    end)
+  end
+end
+
 ---@class NotificationOpts
 ---@field repo string
 ---@field all boolean Whether to show all of the notifications including read ones
@@ -1393,6 +1412,7 @@ function M.notifications(opts)
           map("i", cfg.picker_config.mappings.open_in_browser.lhs, open_in_browser())
           map("i", cfg.picker_config.mappings.copy_url.lhs, copy_notification_url)
           map("i", cfg.mappings.notification.read.lhs, mark_notification_read())
+          map("i", cfg.mappings.notification.done.lhs, mark_notification_done())
           return true
         end,
       })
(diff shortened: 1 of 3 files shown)


williambdean:
Thanks for the PR! @Sharn-konet 

Will take a look when I can

lua/octo/pickers/telescope/provider.lua
```diff
@@ -1330,6 +1330,25 @@ local function mark_notification_read()
...
+      gh.run {
+        args = { "api", "--method", "DELETE", url },
+        headers = { "Accept: application/vnd.github.v3.diff" },
+        cb = function(_, stderr)
+          if stderr and not utils.is_blank(stderr) then
+            utils.error(stderr)
+            return
+          end
+        end,
+      }
```
williambdean, line 1347: Could you use the new syntax for api requests

Might look something like this: 
```lua
gh.api.delete {
  "/notifications/threads/{id}", 
  format = { id = selection.thread_id }, 
  opts = { cb = ..., headers = { ... } }, 
}
```

Reference:
https://github.com/pwntester/octo.nvim/discussions/876

williambdean (changes requested):
Thanks for your patience and thanks for the PR @Sharn-konet 

If you have any issues with this change, let me know

lua/octo/pickers/telescope/provider.lua
Sharn-konet: Hey @williambdean, thanks for pointing me in the right direction, had a read through and have updated the code to match the functionality that was present before.

I've overwritten the default success notification as it reported through as an empty notification, figured that will be more annoying than helpful.

Let me know if there's any other changes you'd like me to make.

williambdean (approved):
Thanks for this change @Sharn-konet. Works well for me!

PR #1071, opened 2025-05-08
Make notification mapping match the config

williambdean:

The README mapping was different than the actual mapping in the configuration. This makes them match.

### Checklist


williambdean:
What was intended @tris203 ?

PR #1072, opened 2025-05-11
Fix broken processing for review commit

### LINKED ISSUE #1061: `:Octo review commit` is broken (stuck on loading)
When I run `:Octo review commit` for a in-progress PR review, and pick a specific commit to review, the diffed files don't load and get stuck on loading like below. I would expect it to show the diff for that specific commit and get stuck

Image

- Go to a github repo in nvim
- Start review using `:Octo review`
- Pick a specific commit after running `:Octo review commit` using your Telescope picker

### Tell us your environment

- nvim version: v0.11.0
- octo.nvim commit hash: `2a478f8`
- GitHub CLI version: `gh version 2.70.0`
- OS: MacOS `14.6.1`

I did some work myself to see if older versions of neovim had this issue for me and I found out that I don't have for this commit in the `main` branch: https://github.com/pwntester/octo.nvim/commit/75d1f6a09536e3bffa86a781cdb0bd72da57acc3

But one commit later, I start to see the problem: https://github.com/pwntester/octo.nvim/commit/1d2982f4b7d8953d3314848b84ecfd2fb16c6aac

AhmedNSidd:

This PR fixes a bug with the `:Octo review commit` command. Previously, when this command was ran, the plugin would end up just "stuck" and it wouldn't show any changes for that commit.

Fixes #1061 

### Describe how you did it

I debugged through the plugin and found out that it was the `merge_pages` function in the `pull_request.lua` file that was faulty. This `merge_pages` function was being called by the `get_commit_changed_files` function which gets the single commit for whose changes we want to see. Actually the result from the API call contains a list of a single JSON object.

So it sends this list of a single JSON object (corresponding to our one commit whose changes we want to see) to the `merge_pages` function. But `merge_pages` had a bug where it was not handling this case. What it was expecting instead was a list of pages and each page would contain its own set of JSON objects. But it didn't expect the case where it would just be an array of JSON objects, which is the output format for a single commit review. 

Anyway, I fixed this by just simply merging the key-value into the `out` variable directly. It should be noted that the first case where the results contain an array of an array of JSON objects still needs to be handled (so I didn't remove that logic) because the `get_changed_files` function also calls the `merge_pages` function and it needs that case

### Describe how to verify it

Just review any PR in octo and then use the `:Octo review commit` command and then select a commit. You'll see that it shows the review of the commit successfully.

### Checklist

### DIFF
--- a/lua/octo/model/pull-request.lua
+++ b/lua/octo/model/pull-request.lua
@@ -66,8 +66,19 @@ M.PullRequest = PullRequest
 local function merge_pages(data)
   local out = {}
   for _, page in ipairs(data) do
-    for _, item in ipairs(page) do
-      table.insert(out, item)
+    if type(page) == "table" then
+      -- Check if page is array-like or object-like
+      if #page > 0 then
+        -- Array-like, use ipairs
+        for _, item in ipairs(page) do
+          table.insert(out, item)
+        end
+      else
+        -- Object-like: merge its key/value pairs directly into out
+        for k, v in pairs(page) do
+          out[k] = v
+        end
+      end
     end
   end
   return out


samholmes (approved):
This looks correct to me. I approve.

AhmedNSidd:
@williambdean @pwntester I wasn't / am not sure there was anything else that needed to be done in order for this PR to be merged (and the other PRs that I opened after this that are dependent on this afterwards). Since these are bug fixes for already documented features, I didn't consider it necessary to update the documentation, and the tests that I ran seem to be passing. Let me know if there's something more I need to do.

williambdean:
Can you consolidate to a single PR. It will be easier to review

AhmedNSidd:
@williambdean Done. I'll close this PR and the other ones in favor of the consolidated PR at #1077

williambdean:
Thanks!

PR #1079, opened 2025-05-19
chore: use function declaration syntax

PeterCardenas:

With a function assignment syntax (`local x = function`), go to definition will show two results when there really only is one. With a function declaration (`local function x()`), go to definition will only show one result.

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -644,7 +644,7 @@ function M.setup()
         --- Get the description of a label
         --- @param search string
         --- @return table label_info
-        local get_label_info = function(opts)
+        local function get_label_info(opts)
           local item = gh.label.list {
             json = "name,description",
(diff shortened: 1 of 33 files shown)


williambdean:
Awesome. Do you know if this can be enforced with pre-commit?

PeterCardenas:
@williambdean i looked into a bit ago and at the time, there were not linters that specifically enforced this, but there definitely should be

PeterCardenas:
also this should probably just be fixed within `lua_ls` but this is easier 🤷

williambdean:
Yeah, it seems upstream. But if it makes it easier to developer right now, that's still a plus. 

So shall we merge?

PeterCardenas:
let's do it :)

williambdean (approved):

PR #1083, opened 2025-05-20
fix: go to file functionality

Danthewaann:

This PR fixes an issue that was accidentally introduced in https://github.com/pwntester/octo.nvim/pull/538 where the `BufEnter` autocmd broke the existing go to file functionality via `gf`. We should only setup this autocmd if the `use_local_fs` option is set to `true`, as this autocmd was intended to be used only if that option was enabled.

### Describe how you did it

Did a bit manual git bisecting to find the commit that introduced the problem.

### Describe how to verify it

Open up a review for a PR and do `gf` to go the the file. If `use_local_fs` is enabled then `gf` won't work, but I think that might be the point.

### Checklist

### DIFF
--- a/lua/octo/autocmds.lua
+++ b/lua/octo/autocmds.lua
@@ -18,16 +18,18 @@ function M.setup()
     end,
   })
 
-  define({ "BufEnter" }, {
-    group = "octo_autocmds",
-    pattern = { "*" },
-    callback = function()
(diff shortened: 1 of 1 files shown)


williambdean:
Will merge this in @Danthewaann 

Thanks for the contribution!

Issue #1066, opened 2025-04-28
Cannot list PRs for repos hosted on GHES host


matawed:
### Describe what happened and what your expected to happen
`:Octo pr list` 

results in an error and no prs are listed.

![Image](https://github.com/user-attachments/assets/e7de37ea-8d1f-4aa7-b9df-7ebfc2e7f5f4)

However, if I do 
`!gh pr list` inside nvim I get a list

![Image](https://github.com/user-attachments/assets/190b8271-a6b9-48d4-90dc-62acb8c18503)

Also I copied the gh command the plugin tries to run, it works in a terminal

```bash
gh api graphql --paginate --jq . -f 'query=query($endCursor: String) {  
  repository(owner: "swh", name: "rpp") {
    pullRequests(first: 100, after: $endCursor, states:OPEN,, orderBy: {field: CREATED_AT, direction: DESC}) {
      nodes {
        __typename
        number
        title
        url
        repository { nameWithOwner }
        headRefName
        isDraft
        state
      }
      pageInfo {
        hasNextPage
        endCursor
      }
    }
  }
}' -H 'Accept: application/vnd.github.v3+json;application/vnd.github.squirrel-girl-preview+json;application/vnd.github.comfort-fade-preview+json;application/vnd.github.bane-preview+json' --hostname XXXXXXXXXXXX | jq -r ".data.repository.pullRequests.nodes[0]"
```

Hostname in terminal was also copied from a print statement during lua code run, so this is definitely the same.

This cmd yields a result 

![Image](https://github.com/user-attachments/assets/1b3747a0-bed0-4bca-935b-758cdbf422fc)

And on shell level I haven't any enviroment variable. 

![Image](https://github.com/user-attachments/assets/7ffb2fb0-e0e2-46a3-923d-6f4dbc412988)

### How to reproduce it (as minimally and precisely as possible)

see above

### Tell us your environment

- nvim version: v0.10.4
- octo.nvim commit hash:   2a478f8b264a69f71bbff81b3affa2b1e1178b92
- GitHub CLI version: 2.69.0
- OS: Ubuntu 24.04
- GHES: 3.14.10

### Anything else we need to know?

I also tried so configure `github_hostname`  when configuring the plugin, but as you can see above, even without it seems that hostname is correctly deduced.

Also I mean this [code](https://github.com/pwntester/octo.nvim/blob/master/lua/octo/gh/init.lua#L48-L77) run fine.

matawed:
[...]

PR #1084, opened 2025-05-20
fix(env): add DBUS env var to enable accessing token in keyring

matawed:

Use environment variable `DBUS_SESSION_BUS_ADDRESS`  to access token in keyring otherwise every `gh` will fail in Octo.

Fixes: #1066 

### Describe how you did it

Inject environment variable `DBUS_SESSION_BUS_ADDRESS` when running `gh`  commands

### Describe how to verify it

See #1066 for details

### Checklist


matawed:
I just committed without having a look at contribution guideline, that's why it was in draft state. If you think the PR is fine, we can merge it. 

At least for me #1066 is fixed, but I have only verified it on Ubuntu 24.04. However, since only an environment variable is injected to `gh` runs it should at least not affect any environment where it's not set.

williambdean:
Thanks for the PR @matawed !

PR #1087, opened 2025-05-26
feat(review): add treesitter highlighting to thread comments

### LINKED ISSUE #602: Use treesitter highlighting in code blocks
## Issue Description

Type:  *feature request*

Code blocks are not very readable in comments/pr descriptions

image

The native markdown parser supports syntax highlighting inside code blocks, we may be able to configure something to support this!

image

### Tell us your environment
```
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713773202
Run "nvim -V1 -v" for more info
```

I have not yet deepdived into my ts config, but I believe default ops ought to do this in recent versions of nvim, let me know if there are reproducibility issues

PeterCardenas:

Highlights the diff hunk selected by review comments with treesitter, similar to the github UI. This maintains the background color already sets by highlights, but adds the additional ones given by treesitter.

Closes #602

### Describe how you did it

The treesitter parser for markdown does have the ability to highlight injected code blocks, but I didn't want to change the way we were visualizing the code block into something worse to accomodate the treesitter parser. So I query the highlights for the code block using `vim.treesitter.get_string_parser`, calculate the content and highlight combinations and convert these to content to be consumed by the existing thread_snippet renderer.

A note: the code is fairly complex, but I think it's well worth it. Unfortunately it doesn't seem like we get the highlight combinations for free with any other method, but with the current method of pre-calculating the highlight and content tuples, it seems like we are doing a lot of work that can be done by just stacking separate extmarks on top of each other. Looking at the [snacks.nvim](https://github.com/Aietes/snacks.nvim/blob/4a6ed6912e5b5cabeb3c3dbbecaedf7a1c743256/lua/snacks/picker/util/highlight.lua#L126) implementation, only the extmark boundaries are calculated and we just apply them directly to the buffer. We will still have to do some post processing to make sure that the ext marks only apply to the content inside the border of the preview and at the correct lines and columns. However, to make this change as self contained as possible, I prefer the current approach so that we don't have to do line counting and calculating in places unrelated to the diff highlighting to make sure to apply to extmarks correctly.

### Describe how to verify it

Open a PR with review comments with code snippets.

### Checklist

  - no documentation updates needed


lua/octo/ui/writers.lua
```diff
@@ -976,7 +980,236 @@ local function get_lnum_chunks(opts)
...
+
+---@param bufnr integer
+---@param diffhunk string
+---@param diffhunk_lang string?
+---@param start_line integer?
+---@param comment_start integer
+---@param comment_end integer
+---@param comment_side "RIGHT" | "LEFT"
+---@return integer
+---@return integer
```
williambdean: Why are there two returns here?  


And this is an example code block :

```lua
--- Comment
local gh = require "octo.gh"
local utils = require "octo.utils"

gh.pr.list {
    opts = {
        cb = gh.create_callback {
            failure = vim.notify,
        }
    }
}
```

williambdean (commented):
Think that this is good to go. Would my example comment be an example of how to use this feature?

lua/octo/ui/writers.lua
PeterCardenas: I updated this to make it more clear that it is a tuple. And yes that is an example! You should be able to see it highlighted properly in octo.

PeterCardenas:
Another note, I just added a commit that should partially solve the issue of highlights overriding each other improperly, but it doesn't seem to fully work. It's possible that we need to be a little smarter about how we handle the ordering of the highlight groups, but would have to look in further. A particular example I found in the wild was the a function call being highlighted as a variable. In any case, it seems to work 98% of the time for my use cases, so just leaving this as a note.

williambdean:
Thanks @PeterCardenas !

PR #1091, opened 2025-05-29
fix(pull request): do not show comment code block when empty

PeterCardenas:

Adding a file comment still shows the thread snippet because the diffHunk ends up being an empty string. We don't handle this case since we end up showing an empty box with a bunch of new lines.

### Describe how to verify it

Add a file comment to a PR and view it in octo.

### Checklist


williambdean (approved):
LGTM

PR #1095, opened 2025-06-09
Add basic support for issue types

williambdean:

This adds issue type to: 

- details
- timeline items

### Checklist


williambdean:
Would you be able to take a look at this? @PeterCardenas

PeterCardenas:
sorry for late review here, just checked this out and looks like we're [missing a few colors](https://docs.github.com/en/graphql/reference/enums#issuetypecolor) which is causing rendering the issue for these unknown colors to crash. A repro is this issue: https://github.com/ghostty-org/ghostty/issues/7557. was the reason we wanted to hard code colors was to match github?

williambdean:
Thanks. I will follow up on it

PR #1096, opened 2025-06-09
feat: open release from notification

PeterCardenas:

Allows viewing releases in the telescope notifications picker. I still need alternative UI to clear releases from my notifications, and this allows me to use octo for that.

### Describe how you did it

There is a decent amount of special casing to allow for this. For one, the notification URL for the release gives the release ID, which from my investigation, was not query-able with the graphql API. Thus, I created a helper that converts the release ID to the tag name. I used this to be able to open the release buffer, which is the same one I use for the preview. I also had to change the selection logic to do this conversion to open up the buffer unless it was already done and cached already by rendering the notification preview.

### Describe how to verify it

This is a bit hard to verify. It's been working in my fork for a while (which admittedly has other modifications). I would guess this would need to be tested by either waiting for a release to pop up in notifications or trigger it oneself by creating a repo, subscribing to it, and then releasing.

### Checklist

  - none needed (no config, keymap or command changes)

### DIFF
--- a/lua/octo/pickers/telescope/previewers.lua
+++ b/lua/octo/pickers/telescope/previewers.lua
@@ -7,6 +7,7 @@ local utils = require "octo.utils"
 local previewers = require "telescope.previewers"
 local pv_utils = require "telescope.previewers.utils"
 local ts_utils = require "telescope.utils"
+local release = require "octo.release"
 local defaulter = ts_utils.make_default_callable
 local workflow_runs_previewer = require("octo.workflow_runs").previewer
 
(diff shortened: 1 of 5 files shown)


lua/octo/pickers/telescope/previewers.lua
```diff
@@ -140,10 +148,37 @@ local notification = defaulter(function(opts)
...
 
       if self.state.bufname ~= entry.value or vim.api.nvim_buf_line_count(bufnr) == 1 then
-        local number = entry.value
+        local number = entry.value ---@type string
         local owner, name = utils.split_repo(entry.repo)
 
+        ---@type string, table<string, string>, string, fun(obj: any, bufnr: integer): nil
         local query, fields, jq, preview
+
+        local function fetch_and_preview()
```
williambdean: Can we pass the parameters here instead of relying on scope

williambdean (commented):
Thanks for the work on this. Will do deeper dive later

williambdean:
I don't have any notifications for releases at the moment... But the logic seems fine to me. Is this good to go; we can iterate on it later if needed.

PeterCardenas:
this has been working well for me so there's at least that 😅, addressed your comment, should be good to go!

Issue #652, opened 2024-10-30
`Octo search` default repo


wd60622:
No description provided.

PR #1112, opened 2025-06-24
fix: allow specifying repo flag

PeterCardenas:

Allows for specifying `repo:` filters for `search` commands. Also allows for not specifying the repo at all and defaulting to the current repo while still using filters.

### Describe how to verify it

`:Octo pr search repo:neovim/neovim is:open`

`:Octo pr search repo:neovim/neovim`

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -960,6 +960,10 @@ function M.process_varargs(repo, ...)
   local args = table.pack(...)
   if utils.is_blank(repo) then
     repo = utils.get_remote_name()
+  elseif string.find(repo, ":") or string.find(repo, "=") then
+    args.n = args.n + 1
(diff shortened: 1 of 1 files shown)


williambdean:
I thought the current repo was already the default behavior? 

Reference: https://github.com/pwntester/octo.nvim/issues/652

mosheavni:
just use `:Octo search is:pr is:open repo:neovim/neovim`.

PeterCardenas:
@williambdean It is if you don't specify any filters, but I've found that you can't specify the repo filter at all because it gets overridden by the logic where it uses the first parameter as the repo.

@mosheavni I know that works, but just fixing the broken `Octo pr search` and `Octo issue search` commands here.

williambdean:
Makes sense. Thanks!

PR #1113, opened 2025-06-24
feat: add multiple reviewers/assignees at once

PeterCardenas:

Allows for adding multiple reviewers/assignees at once. Implementation is a bit hacky in that there is a request for the user id for each user, when it probably could be done in a single request. Though, I don't usually think there's so many reviewers that it matters.

### Describe how to verify it

`Octo reviewer add user1 user2`

`Octo assignee add user1 user2`

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -847,16 +847,16 @@ function M.setup()
       end,
     },
     assignee = {
-      add = function(login)
-        M.add_user("assignee", login)
+      add = function(...)
+        M.add_user("assignee", { ... })
       end,
       remove = function(login)
         M.remove_assignee(login)
       end,
     },
     reviewer = {
-      add = function(login)
-        M.add_user("reviewer", login)
+      add = function(...)
+        M.add_user("reviewer", { ... })
       end,
     },
     reaction = {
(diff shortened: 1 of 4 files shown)


lua/octo/commands.lua
```diff
@@ -2266,12 +2268,14 @@ function M.add_user(subject, login)
     utils.error "Cannot get issue/pr id"
   end
 
-  local function cb(user_id)
-    local query
+  ---@param user_ids string[]
+  local function cb(user_ids)
+    local query ---@type string
+    local user_ids_str = '["' .. table.concat(user_ids, '", "') .. '"]'
```
williambdean: If the gh.run get refactored to `gh.api.graph` then this formatting is done automatically. 

```lua
local gh = require "octo.gh"

gh.api.graphql {
  query = query, 
  f = {
    users = { "first_login", "second_login" },
  },
}
```
but needs the query to be parametrized as well. 

Up for that migration with this or could you make an issue for that

lua/octo/gh/mutations.lua
```diff
@@ -878,7 +878,7 @@ mutation {
 -- https://docs.github.com/en/graphql/reference/mutations#addassigneestoassignable
 M.add_assignees = [[
 mutation {
```
williambdean: would have to parameterize the query here

williambdean (approved):
Looks good but suggestion. Are you open to do now or come back to?

PeterCardenas:
much cleaner! just updated in the last commit

Issue #646, opened 2024-10-21
Icons in Review Submit


wd60622:
### Discussed in https://github.com/pwntester/octo.nvim/discussions/504

<div type='discussions-op-text'>

<sup>Originally posted by **josephcrawfordSRH** February 29, 2024</sup>
I tend to put icons in my PR approvals so that we can see at a quick glance that I approved it, I tried to use `:books` as a test but that did not actually put the emoji in place.  I know you can do it with reactions, is there a way to do it in the message for approving a PR?</div>

PR #1120, opened 2025-07-03
fix: make emoji conceal work again

PeterCardenas:

Makes the emoji conceal from the syntax file work again. Maybe we should add even more emojis, but just want these ones to work for now.

### Describe how to verify it

open an octo buffer that has `:heart:` and see that it gets concealed with a heart emoji.


PeterCardenas:
another option is to use this plugin: https://github.com/Freed-Wu/conceal.vim, that way we unload the responsibility of this functionality to the plugin

williambdean:
Nice. Giving it a try now. :+1: :heart:

It doesn't seem to work on my end. The `:heart:` just goes away. Maybe something to do with my color schema.

PeterCardenas:
seems like all of them work except for the heart emoji which becomes this weird unicode heart? i think i'll just remove it
:heart: :+1: :see_no_evil: :laughing: :thinking_face: :thinking: :ok_hand: :upside_down_face: :grimacing: :rocket: :blush: :tada: :shrug: :man_shrugging: :face_palm: :man_facepalmin: :cowboy_hat_face: :worried: :scream: :white_check_mark:

oh, looks like the heart emoji is multiple code points and the `conceal` in `matchadd` only takes the first character which parses as ❤ instead of ❤️

williambdean:
Does this solve this: 
https://github.com/pwntester/octo.nvim/issues/646

williambdean:
This doesn't work for me with my current setup. Let's merge this. Maybe there is some other issue

PeterCardenas:
I think it does solve it, but testing the current state now, and there might be performance issues. will need to test it out more

PeterCardenas:
i'm gonna close this PR since I don't think this solution is worth it due to the performance issues. The alternative would be to use this plugin: https://github.com/ronisbr/emojify.nvim or vendor the logic into this plugin. lmk which you prefer

PR #1093, opened 2025-05-30
feat(review): add support for showing diff for suggestions

PeterCardenas:
### Describe what this PR does / why we need it

Currently we do not visualize the diff from the suggestion. What this PR does is change the diff hunk to take the suggestion into account and visualize it properly.

TODOs:
- [x] Create a diff hunk for the suggestion
- [ ] Include all the lines in the suggestion diff hunk in the preview
  - Visualization currently works for single line suggestions
- [ ] Show the original diff hunk and the suggestion diff hunk separately

### Does this pull request fix one issue?

closes #111
closes #109

### Describe how you did it

By default, the diff hunk for code suggestions is a just a bunch of additions. So we need to change it so that the selected lines become deletions and the suggestion lines become additions.

### Describe how to verify it

Make a suggestion and open the pull request in octo.nvim

### Checklist

- [ ] Passing tests and linting standards
- [ ] Documentation updates in README.md and doc/octo.txt

PR #1121, opened 2025-07-03
feat: update DiffChange and DiffText to be more like GitHub

PeterCardenas:
Instead of using a neutral color for DiffChanged and DiffText highlights for lines where the contents within a line have changed,
we use red on the left side and green on the right side. This is more in line with how GitHub colors its diffs and I think it makes it easier
to detect changes.

didn't find any

### Describe how to verify it

opened [this PR](https://github.com/pwntester/octo.nvim/pull/1093) in octo and see before and after

before

image

after

image

### Checklist


mmirus:
:wave: this broke my own octo review diff highlight customization (I have it set up to look like diffview.nvim's enhanced diff highlighting). After some experimentation, I was able to override the namespaced highlights added in this PR to get my preferred highlighting back.

I suspect few people are doing this kind of customization, but I wanted to mention the problem as a data point in case making this optional or customizable is ever under consideration.

Thanks for your work on octo!

PeterCardenas:
@mmirus can you share a picture of the customization you have? I'll do some due diligence to see if I can avoid overrriding the existing highlights and create Octo specific highlights. This also broke the highlights in review thread comments so will need to look into that as well.

PeterCardenas:
oh i see it now, yea that should be the default behavior. i'll try to look at their code to duplicate the highlighting behavior while considering the above

mmirus:
Sure thing. As I mentioned, I'm reproducing the enhanced diff highlighting feature from diffview.nvim ([description](https://github.com/sindrets/diffview.nvim/blob/main/doc/diffview.txt#L396-L415)).

Here are a couple examples of how Octo displayed a review diff before and after this PR.

| Before    | After |
| -------- | ------- |
| image  | image    |
| image | image     |

mmirus:
> oh i see it now, yea that should be the default behavior. i'll try to look at their code to duplicate the highlighting behavior while considering the above

Responded before I saw your follow-up. :grin: Thanks for checking into this! Let me know if there's anything I can do to help.

If it's useful, diffview basically just sets window-local highlights for the left and right diff windows that link existing hl groups to their groups [here](https://github.com/sindrets/diffview.nvim/blob/4516612fe98ff56ae0415a259ff6361a89419b0a/lua/diffview/scene/views/standard/standard_view.lua#L68-L82). The also make an adjustment to their highlighting based on the enhanced diff option [here](https://github.com/sindrets/diffview.nvim/blob/4516612fe98ff56ae0415a259ff6361a89419b0a/lua/diffview/hl.lua#L479-L481).

It's pretty similar to what you did in this PR.

PR #1022, opened 2025-04-03
Add `Octo comment reply` and `<leader>cr` mapping

williambdean:
### Describe what this PR does / why we need it

Support for quote replying the comment that the cursor is currently on.

PR #1125, opened 2025-07-04
Add `Octo comment reply` and `<leader>cr` mapping

storopoli:

Support for quote replying the comment that the cursor is currently on.

Closes the old draft PR #1022.

### Describe how you did it

Ok this f**** thing was tricky. I've spent more than 1 hour fiddling around.

What DOES NOT work? You can't add more `\n` (empty lines) in the `format_reply()` because it goes under the `utils.trim()`. You can't change the `utils.trim()` because it breaks everything and you get out of bounds folds. Also you can add blindly more lines to the output of the `utils.trim()` because you get your cursor out of the bounding box of the comment and the user will start writing stuff that will never appear in the reply (the reply will be only the quoted content and no user reply).

So the work around was not elegant but works.
After the `format_reply()` gets shoved into the `utils.trim()` in `add_pr_issue_or_review_thread_comment_reply()`, we detect if lines starts with `>` until we find the last quoted one. Then we add two empty lines after that with `vim.api.nvim_feedkeys("\n\n", "n", false)` (we can't add just one because then it becomes part of the quote, so we need 2).

### Describe how to verify it

Go to any PR and reply to a comment.

### Special notes for reviews

If you have a better way to do this please suggest.

This PR is composed of two commits:

1. The original implementation in #1022 with the original author @williambdean but squashed to a single commit
2. My fix to the cursor thing described in the "Describe how you did it" section above.

### Checklist


williambdean:
How would you like to go with this?

storopoli:
We can merge whatever you want. I'd suggest we just squash merge it.
I left the commits separated so that you can easily review what is your original changes and what I've added on top.

storopoli:
@williambdean I've deleted the `merged master` commits and retained the original 2 commits and rebased on `master`. Do you want me to squash these 2 commits into a single one?

williambdean:
> @williambdean I've deleted the `merged master` commits and retained the original 2 commits and rebased on `master`. Do you want me to squash these 2 commits into a single one?

All good. I will squash when merging.

Thanks for the effort on this!

PR #1126, opened 2025-07-07
fix: floating window content offset

PeterCardenas:
The content of the floating window was positioned at the same y offset as the header. This was because the `header` field was not being passed. Made the types prevent this by making it `boolean` instead of `string?`

### Describe how to verify it

`:Octo pr checks` on this PR


williambdean (approved):
@PeterCardenas is on a roll 🚀 
Thanks for the PR!

PR #1135, opened 2025-07-15
chore: improve typing of OctoBuffer.node

PeterCardenas:
Discourage use of `OctoBuffer.node` in favor of `OctoBuffer:issue()`, `OctoBuffer:pull_request()`, `OctoBuffer:repository()`, `OctoBuffer:release()`, and `OctoBuffer:discussion()`.
This removes the need for casting on the `node` property since it is abstracted in a safe manner.

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -150,12 +150,12 @@ function M.setup()
 
         gh.api.graphql {
           query = mutations.reopen_discussion,
-          fields = { discussion_id = buffer.node.id },
+          fields = { discussion_id = buffer:discussion().id },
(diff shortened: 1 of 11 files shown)


PeterCardenas:
that function (`get_current_buffer`) already infers the type of `octo_buffers` to return `OctoBuffer`, so no changes needed there! i converted most usages here to use the methods instead of `buffer.node.<property>`, but for things like `buffer.node.url` and `buffer.node.id` where the field should be shared across most if not all types, i'm still not too sure what to do there. if i think of something, can do in a followup 🙏

PR #1138, opened 2025-07-16
Type hint merge method

williambdean:

Inspired by @PeterCardenas. Maybe you can take a quick look. I am new to these type hints.

Also, how can we expose the octo.config.OctoConfig? That isn't coming up with my LSP but other various do...

### Checklist

### DIFF
--- a/lua/octo/config.lua
+++ b/lua/octo/config.lua
@@ -5,7 +5,7 @@ local M = {}
 ---@alias OctoMappingsList { [string]: table}
 ---@alias OctoPickers "telescope" | "fzf-lua" | "snacks"
 ---@alias OctoSplit "right" | "left"
----@alias MergeMethod "squash" | "rebase" | "commit"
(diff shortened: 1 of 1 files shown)


PeterCardenas:
the type should be able to be used as `OctoConfig`, not `octo.config.OctoConfig`. the type names could probably be more consistent, but don't think it's a huge deal.

lua/octo/config.lua
```diff
@@ -5,6 +5,7 @@ local M = {}
 ---@alias OctoMappingsList { [string]: table}
 ---@alias OctoPickers "telescope" | "fzf-lua" | "snacks"
 ---@alias OctoSplit "right" | "left"
+---@alias MergeMethod "squash" | "rebase" | "commit"
```
PeterCardenas: prefer naming it `OctoMergeMethod` as a sort of namespacing

williambdean:
Thanks for review @PeterCardenas !

PR #1163, opened 2025-09-08
fix(fzf-lua.search): do not block ui when querying

PeterCardenas:

Query for search is blocking for fzf-lua when it doesn't have to be. Made it non-blocking here.


PeterCardenas:
note: the current fzf-lua picker is broken on latest fzf-lua due to breaking changes on their side 😕, but this fixes it for versions before the breaking changes at least

PR #1166, opened 2025-09-08
fix(pr/issue): handle trailing accumulated timeline events

PeterCardenas:

Accumulated events would not show up if they were the last events fetched. In this PR we make sure to "finalize" the accumulated events by treating a "nil" event as the marker that the events finished. Moved the accumulated events logic to its own function to simplify this.


williambdean:
Thanks @PeterCardenas !

PR #1171, opened 2025-09-17
fix: set buftype=nofile for left buffer

Danthewaann:

Some LSPs will attempt to attach to the left buffer when doing a review.

For example, `ruff` will fail to attach since the left buffer isn't a real file and it will continually print out this warning message:

```
Client ruff quit with exit code 2 and signal 0. Check log for errors
```

To prevent LSPs from attaching to the left buffer, we can set the `buftype` option to `nofile` so LSPs won't attempt to attach to this buffer, since no LSP should be able to attach to it anyway.

NONE

### Describe how you did it

Set `buftype=nofile` for the left buffer

### Describe how to verify it

Open up a review and verify that your LSPs don't attach to the left buffer via `:LspInfo`. Also no warning messages should appear.

### Checklist

### DIFF
--- a/lua/octo/reviews/file-entry.lua
+++ b/lua/octo/reviews/file-entry.lua
@@ -317,6 +317,9 @@ function FileEntry:load_buffers(left_winid, right_winid)
     vim.api.nvim_win_set_buf(split.winid, split.bufid)
   end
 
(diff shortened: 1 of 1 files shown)


williambdean:
Thanks @Danthewaann !

PR #1272, opened 2025-11-12
feat: be able to set static search for better performance

jugarpeupv:
It lets you define if the search command is done with static option or not. For me the current behavior without static option makes telescope lag, i think its because everytime you time it refetches the content

With this approach you can set if you want to disable this feature by default in the picker config

### Checklist

### DIFF
--- a/lua/octo/pickers/fzf-lua/pickers/notifications.lua
+++ b/lua/octo/pickers/fzf-lua/pickers/notifications.lua
@@ -115,12 +115,14 @@ return function(opts)
 
   fzf.fzf_exec(get_contents, {
     prompt = picker_utils.get_prompt(opts.prompt_title),
+    ---@diagnostic disable-next-line: assign-type-mismatch
     previewer = previewers.notifications(formatted_notifications),
     fzf_opts = {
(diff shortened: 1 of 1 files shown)


williambdean:
What is your workflow here? How can I check this?

jugarpeupv:
@williambdean so the thing is, without the code from this PR, when i ran "Octo search is:issue is:open" and type in the telescope picker, i could see a lot of lag. It seems like with every keystroke is like refetching the content.

I explored the source code, and found out that there was already an option "static" which disables this refetching content, but it was not exposed. I tested setting this option to true and the lag disappeared. So what i did is simply expose it as picker config

jugarpeupv:
@williambdean done!

jugarpeupv:
@williambdean  i am not sure about the typecheck error, copilot suggests adding ---@diagnostic disable but i am not sure that is a good practice, i am not familiar enough with lua

williambdean:
> @williambdean i am not sure about the typecheck error, copilot suggests adding ---@diagnostic disable but i am not sure that is a good practice, i am not familiar enough with lua

I've just added the disable for the next lines instead of at the top. 

Will give another review!

PR #1299, opened 2025-11-18
feat: use pickers for selection of branches while creating PR

f3nrir92:
Simplifies process of creating Pull Requests: no need to enter branch names manually, now configured picker is used to select branch

NONE

### Checklist


williambdean:
Thanks for the PR @f3nrir92 , will take a look shortly. I've been using vim.ui.select for pickers which some picker plugins hook into, would it make sense to implement in that way. It could be easier to maintain.

f3nrir92:
@williambdean Hi! Completly agree that `vim.ui.select` will be easier to maintain, but:

- separate logic for several pickers is more flexible: e.g. `vim.ui.select` does not provide (or I don't know how) interface to set default input
- we have to implement "branches" picker right within /lua/octo/picker.lua, so it will look inappropriate to the rest already implemented pickers

williambdean:
> separate logic for several pickers is more flexible: e.g. `vim.ui.select` does not provide (or I don't know how) interface to set default input

I've used something like this before: 

```lua
local items = {
  "main", "feat/something-awesome", "fix/bug-123",
}
local default = "main"


vim.ui.select(items, {
  prompt = "Select a branch:",
  format_item = function(item)
    if item == default then
      item = item .. " (default)"
    end

    return item
  end,
}, function(choice)
  if choice == nil then
    choice = default
  end
  print("You selected: " .. choice)
end)
```
> we have to implement "branches" picker right within /lua/octo/picker.lua, so it will look inappropriate to the rest already implemented pickers

There is now a `default` picker that uses `vim.ui.select` after https://github.com/pwntester/octo.nvim/pull/1302 which acts as a fallback if picker doesn't 
have a specific implementation. That can be a route to go as well. 

My concerns are just around maintainability and having to know multiple picker implementations.

f3nrir92:
> I've used something like this before:

I didn't though about it this way =/ But in that case there is no option to cancel input 

> There is now a default picker that uses vim.ui.select after https://github.com/pwntester/octo.nvim/pull/1302 which acts as a fallback if picker doesn't
have a specific implementation. That can be a route to go as well.
>
>My concerns are just around maintainability and having to know multiple picker implementations.

Oh, thanks a lot! I will code support for default picker shortly (until monday maybe).

f3nrir92:
@williambdean sorry for the delay. I've added default picker and fixed some bugs in my code. Could You check the PR again, pls?

f3nrir92:
@williambdean Hi! Hope you are doing well. Can I ask you don't merge this PR yet? I will implement selecting target remote repository before selection of base branch, if there are many remotes.

f3nrir92:
@williambdean 
Hi! Finally I found powers in myself to complete this feature 😬 
Could You please review it?

PR #1324, opened 2025-11-24
chore: add syntax highlighting for graphql queries

PeterCardenas:

steals this devx nicety from https://github.com/folke/snacks.nvim/commit/1d5b12d0c67071320e5572a1f2ac1265904426b3 to add graphql syntax highlighting to graphql strings. Needed a bit of a hackier solution to accomodate for it, but I think it's worth it since it looks nice.

another note: this doesn't behave well with primitive lazy loading (such as lazy loading on `Octo` user command), but the code snippet below seems to work fine

```lua
vim.api.nvim_create_autocmd({ 'BufReadPre' }, {
  group = lazy_load_octo,
  pattern = '*/octo.nvim/lua/octo/gh/*.lua',
  callback = function()
    require('octo')
  end,
})
```

https://github.com/user-attachments/assets/c448627e-925a-4a27-ac13-5c2255a139be

### DIFF
--- a/after/queries/lua/injections.scm
+++ b/after/queries/lua/injections.scm
@@ -1,0 +1,40 @@
+; extends
+
+; allow injections in strings with an inject: <lang> comment
+; need three copies of the same query for:
+; - handling variable member assignment to string
+; - variable member assignment to a string concatenated with other strings
+; - the above but with some edge case i'm not sure about
+; reference: https://github.com/folke/snacks.nvim/commit/1d5b12d0c67071320e5572a1f2ac1265904426b3
+((comment
+  content: (comment_content) @injection.language)
+  (#lua-match? @injection.language "inject%s*:%s*%S+")
+  (#gsub! @injection.language "^%s*inject%s*:%s*(%S+).*" "%1")
+  .
+  (_
+    (_
+      (string
+        content: (string_content) @injection.content))))
+
+((comment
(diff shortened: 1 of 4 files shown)


williambdean:
Awesome 🚀

williambdean:
I have the graphql lsp set up while using graphql files. That is the reason for the file here: 

https://github.com/pwntester/octo.nvim/blob/cf300731ab1846d467b67ecdf1e37a0db4cd8274/.graphqlrc?plain=1#L1-L2

Do you have the setup as well @PeterCardenas? Any idea it is possible bring in LSP support for these sections of code as well? PS, I am working on #1325 in order to do more introspective and help with debugging and development if you are interested.

PeterCardenas:
@williambdean yea it's been a side project i've been wanting to do for a bit. my thought would be to create a [custom graphql code generator](https://github.com/dotansimha/graphql-code-generator) that outputs type safe lua. would be really cool and keep the types in sync. maybe i'll find some time now that i have this week off lol

PeterCardenas:
there's different levels to it, we could generate just the type annotations and read graphql files at runtime, and a more extreme level is that we create a type safe client at run time that gives methods for all the mutations and queries

PeterCardenas:
i think you can also keep the code as is and just put markers for all the graphql strings to get autocomplete inside them

PR #1334, opened 2025-11-24
feat(telescope.notifications): cache notification previews

PeterCardenas:

we fetch notifications each time we hover over them which makes things feel pretty slow. this change makes it so that once we're inside the picker, the notification response used for the notification preview is cached so that it loads instantly when it was already loaded.

this change might be a bit controversial, as it means that `:Telescope resume` will not reload the notifications which means the data might be stale. In this case, we'd probably add some sort of "staleness" check to refetch, or refetch in the background and then replace it. However, in this state, I do already find it pretty useful.

### DIFF
--- a/lua/octo/notifications.lua
+++ b/lua/octo/notifications.lua
@@ -63,17 +63,16 @@ function M.copy_notification_url(notification)
   }
 end
 
----@param bufnr integer
 ---@param owner string
 ---@param name string
 ---@param number string
 ---@param kind string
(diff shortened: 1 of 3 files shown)


williambdean:
Would it make sense to add this for any of the other pickers?

PeterCardenas:
it would make sense for all of them actually, this one just bugged me the most since i probably use it the most

williambdean:
> it would make sense for all of them actually, this one just bugged me the most since i probably use it the most

But the preview information is just the summary and body so don't think there is much harm there.

Issue #951, opened 2025-03-15
Add missing timelineItems


wd60622:
Various timelineItems are missing

PR #1345, opened 2025-11-26
refactor: use text builder pattern

williambdean:

```lua
local TextChunkBuilder = require "octo.ui.text-chunk-builder"

local bufnr = vim.api.nvim_get_current_buf()
TextChunkBuilder:new()
    :text("Hello ", "Comment")
    :text("World!", "String")
    :write_event(bufnr)
```

### Checklist

### DIFF
--- a/lua/octo/pickers/fzf-lua/pickers/notifications.lua
+++ b/lua/octo/pickers/fzf-lua/pickers/notifications.lua
@@ -115,12 +115,14 @@ return function(opts)
 
   fzf.fzf_exec(get_contents, {
     prompt = picker_utils.get_prompt(opts.prompt_title),
+    ---@diagnostic disable-next-line: assign-type-mismatch
     previewer = previewers.notifications(formatted_notifications),
     fzf_opts = {
       ["--no-multi"] = "", -- TODO this can support multi, probably
       ["--header"] = opts.results_title,
       ["--info"] = "default",
     },
+    ---@diagnostic disable-next-line: missing-fields
     winopts = {
       title = opts.window_title or "Notifications",
       title_pos = "center",


williambdean:
Any thoughts on a refactor like this @PeterCardenas ?

PeterCardenas:
interesting idea, definitely makes the code more concise, almost feels like ui library

PeterCardenas:
one potential idea i had as well was to allow making the virtual text real text so that it's easily copiable, having a library would def make that easier

williambdean:
What does the lua api for that look like?

Also, would a change like this make that change (and others you might have) easier or harder?

williambdean:
> one potential idea i had as well was to allow making the virtual text real text so that it's easily copiable, having a library would def make that easier

I always use tmux copy for this 😀

PeterCardenas:
> What does the lua api for that look like?

it'd probably be switching between nvim_buf_write_text and the extmark api. and potentially capturing ModeChanged autocmd to automatically switch between the two but that might not work

> Also, would a change like this make that change (and others you might have) easier or harder?

having any abstraction would make things easier :) but probably can abstract it further in some ways i'm not thinking of right now

> I always use tmux copy for this

i do too! but it has its limitations. this would make it feel more native to vim

williambdean:
Gotcha. Shall I change the name to TextChunkBuilder or something similar to be more generic. Then we can come back and add more functionality later?

PeterCardenas:
yea that would be great :)

williambdean:
Well I've generalized the name and refactored a bit. Let me know what you think!

williambdean:
Will help with #951

PR #1301, opened 2025-11-18
feat: support plural forms and hostname in URLs

ericboehs:
I clauded this PR this morning when I noticed `:Octo` and `:e octo://` commands weren't working as expected. The implementations in the previous PRs are fine. This is just additional features I felt were missing (enterprise hostname and plural).

---

This PR adds two major improvements to URL handling in octo.nvim:

## 1. Support plural forms in octo:// URLs

Users can now use both singular and plural forms when working with octo:// URLs:

```vim
:e octo://owner/repo/issues/123   " plural form works now
:e octo://owner/repo/issue/123    " singular form still works
:e octo://owner/repo/pulls/456    " plural form works now
:e octo://owner/repo/pull/456     " singular form still works
```

Plural forms are automatically normalized to singular internally.

## 2. Support hostname in regular GitHub URLs

The `:Octo` command now extracts and respects hostnames from full GitHub URLs, making it seamless to work with GitHub Enterprise instances:

```vim
" GitHub.com URLs (existing behavior)
:Octo https://github.com/pwntester/octo.nvim/issues/12

" GitHub Enterprise URLs (new feature!)
:Octo https://ghe.example.com/owner/repo/issues/456
:Octo https://ghe.example.com/owner/repo/pull/789
```

The hostname is automatically detected and passed through to the GitHub CLI API.

## Implementation Details

- Updated `URL_ISSUE_PATTERN` and `URL_RELEASE_PATTERN` to capture hostname
- Modified `parse_url()` to return `(hostname, repo, number, kind)` with hostname as the first parameter for consistency
- Added plural form normalization in `M.load_buffer()`
- Updated all `parse_url()` call sites to use the new return signature
- Documentation updated in both README.md and doc/octo.txt

## Examples

```vim
:e octo://owner/repo/issues/123
:e octo://ghe.com/owner/repo/issue/456
:Octo https://ghe.example.com/owner/repo/pull/789
```

## Testing

All features have been tested with both GitHub.com and GitHub Enterprise instances.

## Note for lazy.nvim users

When using lazy.nvim, add `"BufReadCmd octo://*"` to the event list to ensure the plugin loads when opening octo:// URLs directly:

```lua
{
  "pwntester/octo.nvim",
  cmd = "Octo",
  event = { "VeryLazy", "BufReadCmd octo://*" },
  ...
}
```

PR #1349, opened 2025-12-01
fix: cursor hold on issue/pr urls

PeterCardenas:

Cursor hold previews on pr/issue urls were broken from [this change](https://github.com/pwntester/octo.nvim/pull/1301), so this PR handles the extra capture group for the hostname.


williambdean:
Thanks @PeterCardenas !

Any idea how we can make elements like this more robust?

PeterCardenas:
unit tests would likely cover this

PR #1371, opened 2025-12-04
fix: locally set file window options

igorlfs:

When an option is set globally (the default with `nvim_set_option_value`), other windows inherit its value.

If I use a picker to create a new tab with the file I'm reviewing, by that logic, it will inherit all those window options! Most notably, it affects the `foldmethod` option, making me unable to use treesitter's folding on the new tab.

There are some other usages of `nvim_set_option_value` as well. They _probably_ should use `"local"` as well, though I personally haven't run into issues with them (hence I decided to not touch them)

None

### Describe how you did it

By adding a parameter.

### Describe how to verify it

- `Octo review browse`
- `Fzflua files`
- Pick file (`<C-t>` to open on a new tab)

### Checklist

- [ ] Documentation updates in README.md and doc/octo.txt **N/A**

### DIFF
--- a/lua/octo/reviews/file-entry.lua
+++ b/lua/octo/reviews/file-entry.lua
@@ -596,7 +596,7 @@ end
 function M._configure_windows(left_winid, right_winid)
   for _, id in ipairs { left_winid, right_winid } do
     for k, v in pairs(FileEntry.winopts) do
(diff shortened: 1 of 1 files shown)


williambdean (approved):

williambdean:
Thanks for the PR! @igorlfs

Issue #1389, opened 2025-12-06
Navigate to comment from notification


williambdean:
The notification payload can provide information about the comment

Some steps needed to implement this feature:

- [x] Pull in databaseId for IssueComment
- [x] Navigate to a comment based on the databaseId
- [ ] Parse the comment databaseId from the notification payload 
- [ ] Bring together

PR #1391, opened 2025-12-08
feat: navigate to comment based on id

williambdean:

Helper for https://github.com/pwntester/octo.nvim/issues/1389

### Checklist

### DIFF
--- a/lua/octo/model/comment-metadata.lua
+++ b/lua/octo/model/comment-metadata.lua
@@ -2,6 +2,7 @@ local M = {}
 
 ---@class CommentMetadata
 ---@field id string
+---@field databaseId integer
 ---@field author string
 ---@field savedBody string
 ---@field body string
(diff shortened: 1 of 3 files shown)


williambdean:
This is an example comment

williambdean:
This script can be used to navigate to a specific comment in a buffer:

```lua
local utils = require "octo.utils"

vim.api.nvim_create_user_command(
  "Navigate",
  function()
    local buffer = utils.get_current_buffer()

    if buffer == nil then
      utils.error "No octo buffer found"
      return
    end

    local comments = {}
    for _, comment in ipairs(buffer.commentsMetadata) do
      table.insert(comments, comment)
    end

    if #comments == 0 then
      utils.error "No comments found in this buffer"
      return
    end

    vim.ui.select(comments, {
        prompt = "Select a comment to navigate to:",
        format_item = function(comment)
          return comment.body
        end,

      },
      function(selected_comment)
        if selected_comment == nil then
          return
        end

        buffer:navigate_to_comment { databaseId = selected_comment.databaseId }
      end
    )
  end, {})
```

PR #1413, opened 2026-01-06
feat(Snacks): rich user picker

maisieccino:
Improves the Snack user picker with a rich user preview.

Sometimes it's hard to know which users you're adding for review. This change shows the user's name, pronouns and profile picture to make it easier!

Embedded images are included with `Snacks` for supported terminals. [Documentation](https://github.com/folke/snacks.nvim/blob/main/docs/image.md)

image

NONE

### Describe how you did it
Updated the GraphQL user queries, before using a Markdown template string to render the data.

### Describe how to verify it
`:Octo reviewer add` to launch the prompt, with `opts.picker = "snacks"`

### Special notes for reviews
Please let me know if the format should change or any details could be added
from user profiles, I've tried to keep it somewhat minimal.

I'll also try and add this for teams too in a future PR!

### Checklist

### DIFF
--- a/lua/octo/commands.lua
+++ b/lua/octo/commands.lua
@@ -2726,11 +2726,11 @@ function M.add_user(subject, logins)
     end
     cb(user_ids)
   else
-    picker.users(function(user_id)
-      if type(user_id) == "string" then
-        cb { user_id }
+    picker.users(function(user_ids)
+      if type(user_ids) == "string" then
+        cb { user_ids }
       else
-        cb(user_id)
+        cb(user_ids)
       end
     end)
   end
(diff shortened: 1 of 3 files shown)


williambdean:
Screenshot 2026-03-04 at 4 56 44 PM


I'm giving this a try. Any thoughts why it doesn't seem formatted correctly?

maisieccino:
> Screenshot 2026-03-04 at 4 56 44 PM
> I'm giving this a try. Any thoughts why it doesn't seem formatted correctly?

hey @williambdean! if I had to guess, think it's because the image isn't getting resized and it's getting cut off. i was able to replicate this too:

image


i guess as well, i think i have a markdown plugin that's formatting for me :thinking:  i'll go away and think about how to make this a bit more portable. might be that it has to use the snacks image API directly.

Issue #1162, opened 2025-09-08
RFC: some default mappings are too opinionated


igorlfs:
Hello,

Recently, I was reviewing a PR when I noticed I couldn't use `<C-e>` to scroll down one of the windows. Taking a look at the docs, I realized it's bound to an action I never used (`copy_sha`). It wasn't the first time I had trouble with Octo's default bindings: all bindings to submit a review also suffer similar issues:
- `<C-a>` is a default for incrementing a number value;
- `<C-m>` is a synonym for `<CR>` (which means `<CR>` is also bound to execute a `comment_review`, this is specially annoying now that these keys are [also mapped in insert mode](https://github.com/pwntester/octo.nvim/pull/847) (i.e., you can't use `<CR>` to insert a new line in insert mode));
- `<C-r>` is `redo` in normal mode and can be used to insert the contents of a register in insert mode.

There are some other `<C-` based mappings (I haven't had trouble with them yet), but overall I think all of them should be avoided to prevent overlapping with built-in functionality. Since there are many, many more mappings that use `<localleader>`, perhaps it could be used as a standard?

For the record, I know I can just disable the built-in mappings via `mappings_disable_default` (which is what I've been doing), my goal with this issue is improving the OOTB experience.

williambdean:
Thanks for flagging. It would be good to mitigate this as much a possible. 

Is your list exhaustive? Would you like to make a PR?

PR #1414, opened 2026-01-07
fix: improve default binds for submit_win

Danthewaann:

The default `submit_win` binds make it difficult to write text in the submit PR review window as typing `<Enter>` (`<C-m>` sends the same keycode as `<Enter>`) in insert mode will submit your review instead of moving the cursor to the next line. Also `<C-a>` and `<C-r>` has uses in insert mode that should be usable in the submit_win. These binds are now only set in `normal` mode for the submit_win.

I also removed the autocmd that added a `BufLeave,BufDelete` event handlers to the submit_win as they cause the window to close prematurely when I accept a completion menu item via `<C-y>` in insert mode.

NONE

### Describe how you did it

I've described what I have done above

### Describe how to verify it

Attempt to submit a PR review and verify that you can accept completion menu items and you can press `<Enter>` in insert mode to move the cursor to the next line in the submit_win

### Checklist

### DIFF
--- a/lua/octo/config.lua
+++ b/lua/octo/config.lua
@@ -449,10 +449,10 @@ function M.get_default_values()
         unresolve_thread = { lhs = "<localleader>rT", desc = "unresolve PR thread" },
       },
       submit_win = {
-        approve_review = { lhs = "<C-a>", desc = "approve review", mode = { "n", "i" } },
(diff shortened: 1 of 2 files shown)


igorlfs:
Related: https://github.com/pwntester/octo.nvim/issues/1162

IMO we should move from `<C-` based mappings altogether

PR #1419, opened 2026-01-12
fix(pr): enter pr checks window on open

PeterCardenas:

We currently set that you don't enter the pr checks window on open. I think this worked before, but doesn't work on nightly, and feels like this is the intended behavior. So explicitly set the window to open here.


williambdean:
thanks Peter!

PR #1423, opened 2026-01-28
feat(labels): allow label selection by prompt text

jugarpeupv:
Enhance label pickers to support selecting a label by typing its name in the prompt, even if no label is selected from the list. This works for both fzf-lua and Telescope pickers. The implementation uses the REST API to fetch the label's node_id directly by name, improving performance and reliability. Error messages are shown if the label cannot be found.

In case the repository has more than 100 labels, the graph qury only returns the first 100 entries

Instead of implementing pagination, which was more complicated, this commit allows for using the prompt input text in telescope/fzf to use it as label, then rest api is used to get the id of the label and then the label is added properly to the issue

Without this code, if the label you want to add is not between the 100 entries then you can not add the label

We have a use case where we add incremental labels in a repository, so that is why we have more than 100 labels

### Describe how to verify it

Create more than 100 labels in a repository and try to add label 101 to it without this commit

### Checklist


lua/octo/commands/project.lua
```diff
@@ -0,0 +1,159 @@
+local gh = require "octo.gh"
+local graphql = require "octo.gh.graphql"
```
williambdean: Not used.

lua/octo/commands/project.lua
```diff
@@ -0,0 +1,159 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+]],
+    project_id
+  )
```
williambdean: Move to queries / mutation

lua/octo/commands/project.lua
```diff
@@ -0,0 +1,159 @@
...
+
+  gh.run {
+    args = { "api", "graphql", "-f", string.format("query=%s", query) },
+    cb = function(output, stderr)
+      if stderr and not utils.is_blank(stderr) then
```
williambdean: Use the gh.api.graphql API

lua/octo/commands/project.lua
```diff
@@ -0,0 +1,159 @@
...
+end
+
+---Update a project field value
+---@param opts { project_id: string, item_id: string, field_id: string, option_id: string }
+---@param cb function Callback on success
+local function update_field_value(opts, cb)
+  local query = string.format(mutations.update_project_v2_item, opts.project_id, opts.item_id, opts.field_id, opts.option_id)
+
+  gh.run {
+    args = { "api", "graphql", "-f", string.format("query=%s", query) },
```
williambdean: Same as above.

williambdean (changes requested):

williambdean:
Will have to run pre-commit

lua/octo/mappings.lua
```diff
@@ -183,6 +183,12 @@ return {
       ["Copy URL"] = commands.issue.url,
       ["Add Type"] = commands.type.add,
       ["Remove Type"] = commands.type.remove,
+      ["Change WS Field"] = function()
+        commands.project.field.set("WS")
+      end,
+      ["Change Status Field"] = function()
+        commands.project.field.set("Status")
+      end,
```
williambdean, line 191: Pre-commit related

lua/octo/ui/writers.lua
```diff
@@ -655,7 +655,7 @@ function M.write_details(bufnr, issue, update, include_status)
   end
   table.insert(details, author_vt)
 
-  add_details_line(details, "Created", issue.createdAt, "date")
+  add_details_line(details, "Creation date", issue.createdAt, "date")
```
williambdean: Where does this come in?

williambdean:
Can you screenshot these changes

jugarpeupv:
I close this issue, since it contains fields only relevant to my use case, sorry for the noise

Issue #601, opened 2024-09-12
Load current branch PR/review data in the background


GuillaumeLagrange:
## Issue Description

Type: *feature request*

Hey guys, I review a lot of code using octo (particularly with `use_local_fs` which is a no brainer to get LSP information during review) and I have a few ideas

### Describe what happened (or what feature you want)
One thing I love about the vs code gh PR plugin is how seamless it is checking out a branch, see the open PR and start reviewing

Since the changes of #492, my workflow is basically
- Checkout a branch
- `Octo pr`
- `Octo review start/resume`
- Do the review, submit

### Describe what you expected to happen
I believe we could improve the first two steps even further with a few loosely linked ideas
1. In the background, regularly fetch PR information for the current branch so that we dont need to wait for the query time
2. Auto refresh the PR buffer on bufenter in addition to this
3. Have a way to enter review diff mode without having to chose between `review start` and `review resume`, basically automatically chose the correct action based on whether a review is already in progress or not
4. Have a way to enter review diff mode without even going through the octo pr buffer (closely linked to 1, if we prefetch all data in the background, review diff is one keypress away)
5. Bonus after 4: have a way to quickly toggle on and off diffmode to test out changes in a file before making comments

### Anything else we need to know?

I will most likely work on almost all of these anyway, but I am testing the waters to see if this could get upstreamed, to see if I need to do it the clean way or not 🤓

Depending on the response I get here, we may break these into different issues to avoid blocking good changes because the rest is quite opinionated.

Let me know what you think, thanks for this awesome plugin ❤️

Issue #97, opened 2021-02-23
Reload active issue/PR buffer with e!


weilbith:
If you are in a big and intensive PR discussion with multiple threads, it would be cool if you could keep the PR buffer open and reload it to see a response by another user (while you have respond in other threads maybe). Not sure if this would work with fetching the autocommand when the use types `:e!`?
This might be related to the idea in #59 to have folds for review conversations in the PR buffer.

pwntester:
You can currently reload the buffer with `Octo issue/pr reload` or just `<c-r>` but I will look into the `e!` idea to make it more vim-ish.

As for including review threads in the main PR buffer. I think we can add a configuration setting to enable it. This should be independent of having folds for octo comments/threads. Please file a different issue for each of these two enhacement requests

weilbith:
> You can currently reload the buffer with Octo issue/pr reload or just <c-r> but I will look into the e! idea to make it more vim-ish.

Thank you very much! Sorry that I missed that. :expressionless:

tami5:
Interesting I didn't know we can do that nice

weilbith:
Sorry mistake. Missed that this is now about the idea to make it more vimish with `:e!` :speak_no_evil:

tami5:
How about step further and let octo.nvim check every 5 seconds async, and update the buffer if current data mismatch the remote data.

weilbith:
> How about step further and let octo.nvim check every 5 seconds async, and update the buffer if current data mismatch the remote data. 

I don't think that's a good idea. As we there is this concept that the buffer is editable and changes get pushed via the API, this could just lead to often to ugly conflicts. I mean there should be no conflicts, but at least some merging strategy would be required.

tami5:
Oh yah increases complexity at this stage, perhaps later down the line or put into  todo list with low priority. @pwntester how about a project and we link issues of that sort to it.

pwntester:
It could make the call and only reload the buffer if its not modified. There can be some cursor position glitches though

pwntester:
LOL, it turns out `e!` already works with current `BufReadCmd` implementation :laugh:

PR #1430, opened 2026-03-04
feat: add polling/auto-refresh for PR and issue buffers

### LINKED ISSUE #815: [Feature Request]: Poll issue and pr buffers for changes
Hello,

I'd like to discuss a feature which I'd like to implement. 

I often leave Octo.nvim in a terminal sitting around in the background. 
It would be nice to get an indication on an open issue or PR buffer that something was updated. 

I'm thinking as a first step, some non-intrusive method can be used to indicate the issue/pr has been updated. 
My first thinking is, just place an asterisk somewhere in the buffer name, so attention is brought to either your tab bar displaying the buffer name or the status bar which shows the buffer name as well. Optionally, `nvim.notify` can be used, but this is only useful for individuals who look at their notifications in a list format. 

After this first step, I think the idea of reloading the buffer with the new data can be explored. 
There is already a reload function for this. 
With reloads comes a bit of complication tho w/r/t pending comments. 

There are a few problematic states here
1. User is currently editing a comment and reload occurs: naive fix: check for VIM mode, if in insert mode, defer the reload to next scheduled interval
2. User has a pending comment: in this case the user may be in normal mode so the above mitigation does not work. Instead, the Octo buffer should keep track of whether in a "pending comment" state and defer reloads until it's not. 

Thoughts?

christopherklint97:

Adds opt-in polling that periodically checks GitHub for remote updates to open issue/PR buffers using a lightweight GraphQL query (only fetches `updatedAt`). When a remote change is detected:
- Buffers **without** local edits are automatically reloaded (cursor position preserved via existing `load_buffer`)
- Buffers **with** unsaved local edits (dirty) are skipped with a user notification
- Users can force-reload dirty buffers with `:Octo poll apply`

New commands: `:Octo poll start/stop/toggle/status/apply`
New config section: `poll.enabled`, `poll.interval`, `poll.notify_on_refresh`, `poll.notify_on_change`

Fixes #815
Related: #601, #97

### Describe how you did it

- **`lua/octo/polling.lua`** (new): Core polling module using `vim.uv.new_timer()` with `vim.schedule_wrap()`. Tracks buffers in a table keyed by `bufnr`. Each tick fires a lightweight GraphQL query per tracked buffer checking only `updatedAt`. Respects the existing `dirty` flag on `titleMetadata`, `bodyMetadata`, and `commentsMetadata` before reloading.
- **`lua/octo/config.lua`**: Added `poll` config defaults (opt-in, 10s interval) with type annotations and validation.
- **`lua/octo/init.lua`**: Hooked `track_buffer()` into `create_buffer()` after `clear_history()` — works for both initial load and reload.
- **`lua/octo/autocmds.lua`**: Added `BufDelete`/`BufWipeout` autocmds for `octo://*` to untrack buffers and auto-stop the timer when no buffers remain.
- **`lua/octo/commands.lua`**: Added `:Octo poll` subcommands.
- **`README.md`** and **`doc/octo.txt`**: Documented config options and commands.

### Describe how to verify it

1. Enable polling in config: `poll = { enabled = true, interval = 5000 }`
2. Open a PR with `:Octo pr`
3. Confirm buffer is tracked: `:Octo poll status`
4. Have someone comment on the PR externally — buffer auto-refreshes within the interval
5. Edit a comment locally (make buffer dirty) — confirm reload is skipped with notification
6. Run `:Octo poll apply` to force-reload dirty buffer with pending changes
7. Close all octo buffers — confirm timer stops (`:Octo poll status`)
8. Test `:Octo poll toggle` to manually start/stop

### Special notes for reviews

- Polling is **opt-in** (`enabled = false` by default) as suggested in #815
- Uses `vim.uv` timers as recommended by @pwntester in #97
- GraphQL query costs ~1 API point per buffer per tick (fetches only `updatedAt` via `issueOrPullRequest` union)
- Timer auto-starts when first buffer is tracked (if `enabled`) and auto-stops when all octo buffers are closed

### Checklist

### DIFF
--- a/lua/octo/polling.lua
+++ b/lua/octo/polling.lua
@@ -1,5 +1,6 @@
 local config = require "octo.config"
 local gh = require "octo.gh"
+local queries = require "octo.gh.queries"
 local uri = require "octo.uri"
 local utils = require "octo.utils"
 local vim = vim
@@ -20,17 +21,6 @@ local tracked_buffers = {}
 
 ---@type uv_timer_t|nil
 local timer = nil
-
-local updated_at_query = [[
-query($owner: String!, $name: String!, $number: Int!) {
-  repository(owner: $owner, name: $name) {
-    issueOrPullRequest(number: $number) {
-      ... on Issue { updatedAt }
-      ... on PullRequest { updatedAt }
-    }
-  }
-}
-]]
 
 ---Check if an OctoBuffer has unsaved local edits
 ---@param octo_buf OctoBuffer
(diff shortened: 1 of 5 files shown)


christopherklint97:
Have been dogfeeding this myself locally and appears to be working quite nice 😄

lua/octo/polling.lua
```diff
@@ -0,0 +1,256 @@
...
+query($owner: String!, $name: String!, $number: Int!) {
+  repository(owner: $owner, name: $name) {
+    issueOrPullRequest(number: $number) {
+      ... on Issue { updatedAt }
+      ... on PullRequest { updatedAt }
+    }
+  }
+}
+]]
+
```
williambdean: Can we add to the queries module

williambdean:
Do you want to add to the commands here:

https://github.com/pwntester/octo.nvim/blob/5ae580df72589f25b775ff2bdacfd7f7be8d63bd/lua/octo/mappings.lua?plain=1#L86-L171

williambdean:
Should be able to run pre-commit run --all-files to solve pre-commit issues.

williambdean:
I've changed the config, done `Octo poll toggle` then `Octo poll status` and have it
saying that it is still on. 

Is the toggle suppose to be global toggle or toggle on current buffer ?

christopherklint97:
Thanks for the review @williambdean! Pushed a fix addressing all feedback:

- Moved the query to queries module
  - The query is now at `queries.updated_at` in `lua/octo/gh/queries.lua`
- Added "Toggle Polling" to both PR and issue options pickers
- `toggle` now also updates `config.values.poll.enabled` at runtime, so toggling off prevents the timer from auto-restarting when new buffers are opened.
- Also did some CI fixes
  - StyLua formatting fixed (expanded multi-arg calls)
  - lua-language-server typecheck fixed: added nil check after `vim.uv.new_timer()`, added `OctoBuffer:get_updated_at()` public method to avoid accessing private `node` field externally

williambdean:
I've found this hitting my GraphQL rate limits. My setup uses tmux so I probably left
some octo buffers open that caused it to keep running.

I'm providing some tooling to check the limits here: https://github.com/pwntester/octo.nvim/pull/1437

PR #1440, opened 2026-03-10
feat: async timeline rendering

williambdean:

Loads meta data first (title, body, etc) then the timeline items. Sound be faster loading times for large PRs and issues.

Maybe you have some thoughts @PeterCardenas !

> [!WARNING]
>
> I am still testing this locally ...

### Checklist

### DIFF
--- a/lua/octo/gh/queries.lua
+++ b/lua/octo/gh/queries.lua
@@ -151,6 +151,239 @@ query($owner: String!, $name: String!, $number: Int!, $endCursor: String) {
   ---@field autoMergeRequest { enabledBy: { login: string }, mergeMethod: string }
 
   -- https://docs.github.com/en/free-pro-team@latest/graphql/reference/objects#pullrequest
+  -- Metadata only version - no timeline items for instant loading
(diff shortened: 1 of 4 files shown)


williambdean:
My one issue with this in practice so far is that the `:Octo reload` becomes very jumpy.
However, it seems to be working fine otherwise. Would appreciate any use and or
feedback.

PeterCardenas:
i'll try to patch it in my fork and report back 🙏 my general thought is that it's difficult to do async right, but maybe it's ok if there's some rough edges

PeterCardenas:
it seems to behave weird with reload and the new polling feature yea, maybe it would be best to just buffer the previous timeline results instead of showing "loading"? also i feel like left aligning the loading text would make it more apparent

PeterCardenas:
also having the timeline items load paginated seems like a regression, feels like it would be better to just load all timeline items at once. or maybe i found some race condition with the pr reload and pollin

williambdean:
Thanks again for the eyes here! Definitely a handful of rough edges here.

PR #1449, opened 2026-03-19
fix auth handling for SSH aliases and multi-account gh logins

thedanielfactor:

This PR fixes Octo auth handling for users with multi-account `gh` setups and SSH alias-based git remotes. It ensures Octo resolves SSH aliases correctly, detects the active `gh` account for the resolved host, and allows switching the active account from inside Neovim.

I did not see an issue like this.  I found it while I was trying to use Octo.

### Describe how you did it

I updated Octo to normalize SSH alias hosts through `ssh_aliases` so aliased remotes like `git@TheDanielFactor:owner/repo.git` resolve to `github.com`. I also changed auth detection to query `gh auth status --json hosts` and use the active account for the resolved host instead of relying on the first parsed login. In addition, I added `:Octo auth status` and `:Octo auth switch [login]` to support inspecting and switching accounts directly from Neovim, and documented the workflow in the README.

### Describe how to verify it

1. Configure an SSH alias in `~/.ssh/config` and set the matching alias in Octo's `ssh_aliases` config.
2. Authenticate multiple GitHub accounts with `gh auth login`.
3. Run `gh auth status` and confirm one account is active.
4. Open Neovim in a repo that uses the SSH alias remote.
5. Run `:Octo auth status` and confirm Octo shows accounts for `github.com`.
6. Run `:Octo auth switch` or `:Octo auth switch <login>` and confirm the active account changes as expected.
7. Confirm Octo commands now work with the selected account.

### Special notes for reviews

This change is aimed at setups where the git remote uses an SSH host alias but `gh` auth is stored against `github.com`. The fix keeps the existing `ssh_aliases` behavior and improves how Octo selects the authenticated account when multiple accounts exist for the same host.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,6 +614,8 @@ If no command is passed, the argument to `Octo` is treated as a URL from where a
 | actions  |                                                   | Lists all available Octo actions                                                                                                                       |
 | search   | <query>                                           | Search GitHub for issues and PRs matching the [query](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests) or Discussions with `is:discussion`|
 | run      | list                                              | List workflow runs                                                                                                                                     |
+| auth     | status                                            | Show authenticated `gh` accounts for the current host                                                                                                  |
+|          | switch [login]                                    | Switch the active `gh` account for the current host                                                                                                    |
 | notification | list                                          | Shows current unread notifications |
 | discussion   | list [repo]                                          | List open discussions for current or specified repo |
 |    | edit <number> [repo] | Edit discussion in current or specified repo |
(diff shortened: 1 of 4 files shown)


williambdean:
I think it is easier to just leverage the existing CLI commands. 

```
$ gh auth
Authenticate gh and git with GitHub

USAGE
  gh auth <command> [flags]

AVAILABLE COMMANDS
  login:         Log in to a GitHub account
  logout:        Log out of a GitHub account
  refresh:       Refresh stored authentication credentials
  setup-git:     Setup git with GitHub CLI
  status:        Display active account and authentication state on each known GitHub host
  switch:        Switch active GitHub account
  token:         Print the authentication token gh uses for a hostname and account
```

No need to change the gh module ... For example, 

```lua
local gh = require "octo.gh"

--- Status to login
gh.auth.status { opts = { cb = gh.create_callback {} } } 

--- Automatically switch accounts
gh.auth.switch { opts = { cb = gh.create_callback {} } }
```

I think wrapper around these would be fine. 

However, in general, the commands are customizable / overideable : https://github.com/pwntester/octo.nvim/discussions/1009

PR #1461, opened 2026-03-31
feat(reviews): allow placing comments outside diff hunks

GeigerJ2:
FULL DISCLOSURE: The actual `src` changes of this PR were done by Claude Code.

However, I had stumbled upon the issue of not being able to add comments outside of hunks, while doing a PR review in nvim using Octo myself. As the changes to my nvim config by Claude Code were applied in my local installation directory at `~/.local/share/nvim/lazy/octo.nvim/lua/octo/reviews/init.lua`, I thought why not contribute back and actually open a PR for it. I hope that's OK.

`pre-commit`, linting, and tests were verified locally. Please don't hesitate to close, if deemed slop. If not, and any changes are required, please let me know, and I'll happily address them (with my very limited Lua knowledge). The current implementation _does_ work for me locally. Many thanks!

Currently, `octo.nvim` blocks placing review comments on lines that fall outside diff hunks with `"Cannot place comments outside diff hunks"`. However, this is a plugin-side restriction — the GitHub GraphQL API ([`AddPullRequestReviewThread`](https://docs.github.com/en/graphql/reference/mutations#addpullrequestreviewthread)) accepts comments on any line of a changed file. The `diff_hunk` is not an input parameter to the [`AddPullRequestReviewThreadInput`](https://docs.github.com/en/graphql/reference/input-objects#addpullrequestreviewthreadinput); it is only returned in the response.

This PR removes that restriction so reviewers can comment on any line visible in the review split, not just lines within diff hunk boundaries.

Fixes #1460

### Describe how you did it

In `lua/octo/reviews/init.lua`, replaced the early return with a placeholder `@@ -0,0 +0,0 @@` for the local thread stub display. The GitHub API receives `path`, `line`, and `side` as before — `diff_hunk` was never sent as input, and GitHub returns its own `diffHunk` in the response.

### Describe how to verify it

1. Open a PR review with `:Octo review`
2. Navigate to a line that is outside any diff hunk (e.g., an unchanged line not shown in the diff context)
3. Try to add a comment on that line
4. **Before**: `"Cannot place comments outside diff hunks"` error
5. **After**: Comment is created and submitted to GitHub successfully

### Special notes for reviews

The change is minimal (single block in `init.lua`). The placeholder is only used for the local thread stub — GitHub populates the actual `diffHunk` field in its response, which is what persists.

No new tests are added, as there is no test infrastructure for the review comment placement flow (which would require mocking the full review layout, file splits, and diff state). All existing tests continue to pass.

### Checklist

- [x] Documentation updates in README.md and doc/octo.txt — N/A, no new user-facing API; this removes an incorrect restriction on existing functionality

### DIFF
--- a/lua/octo/reviews/init.lua
+++ b/lua/octo/reviews/init.lua
@@ -444,8 +444,9 @@ function Review:add_comment(isSuggestion)
     end
   end
   if not diff_hunk then
-    utils.error "Cannot place comments outside diff hunks"
-    return
+    -- The GitHub API does not require comments to be within diff hunks;
+    -- diff_hunk is only used locally and GitHub returns its own in the response.
+    diff_hunk = "@@ -0,0 +0,0 @@"
   end
   if not vim.startswith(diff_hunk, "@@") then
     diff_hunk = "@@ " .. diff_hunk
--- a/lua/octo/reviews/init.lua
+++ b/lua/octo/reviews/init.lua
@@ -446,7 +446,17 @@ function Review:add_comment(isSuggestion)
   if not diff_hunk then
     -- The GitHub API does not require comments to be within diff hunks;
     -- diff_hunk is only used locally and GitHub returns its own in the response.
-    diff_hunk = "@@ -0,0 +0,0 @@"
+    local best_dist = math.huge
+    for i, range in ipairs(comment_ranges) do
+      local dist = math.min(math.abs(range[1] - line1), math.abs(range[2] - line1))
+      if dist < best_dist then
+        best_dist = dist
+        diff_hunk = file.diffhunks[i]
+      end
+    end
+    if not diff_hunk then
+      diff_hunk = "@@ -0,0 +0,0 @@"
+    end
   end
   if not vim.startswith(diff_hunk, "@@") then
     diff_hunk = "@@ " .. diff_hunk
--- a/lua/octo/model/octo-buffer.lua
+++ b/lua/octo/model/octo-buffer.lua
@@ -613,7 +613,7 @@ function OctoBuffer:do_add_new_thread(comment_metadata)
             ---@type octo.mutations.AddPullRequestReviewThread
             local resp = vim.json.decode(output).data.addPullRequestReviewThread
 
-            if utils.is_blank(resp) then
+            if utils.is_blank(resp) or utils.is_blank(resp.thread) then
               utils.error "Failed to create thread"
               return
             end


williambdean:
Thanks for the PR; 

I'm trying it out and getting this: 

```
Error executing vim.schedule lua callback: .../neovim-plugins/octo.nvim/lua/octo/model/octo-buffer.lua:632: attempt to index field 't
hread' (a userdata value)
stack traceback:
        .../neovim-plugins/octo.nvim/lua/octo/model/octo-buffer.lua:632: in function 'success'
        ...ill/GitHub/neovim-plugins/octo.nvim/lua/octo/gh/init.lua:147: in function 'cb'
        ...ill/GitHub/neovim-plugins/octo.nvim/lua/octo/gh/init.lua:259: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

lua/octo/reviews/init.lua
```diff
@@ -444,8 +444,19 @@ function Review:add_comment(isSuggestion)
...
+    -- The GitHub API does not require comments to be within diff hunks;
+    -- diff_hunk is only used locally and GitHub returns its own in the response.
+    local best_dist = math.huge
+    for i, range in ipairs(comment_ranges) do
+      local dist = math.min(math.abs(range[1] - line1), math.abs(range[2] - line1))
+      if dist < best_dist then
+        best_dist = dist
+        diff_hunk = file.diffhunks[i]
+      end
+    end
```
williambdean, line 456: Test comment: Previous behavior. 

This works as expected.

williambdean (changes requested):
Thanks for the PR. I'm trying it out but I am not able to have it work. Commented with
the traceback

GeigerJ2:
Thanks for trying it out and reporting the traceback, @williambdean!

I've since investigated further, and it turns out that when it worked for me locally, my test comments must have landed on lines still within a diff hunk's ±3 context range. I've verified this again, and comments within that range go through fine (see images below), while anything beyond it gets the `thread: null` response.

image

{CEA907B1-B33C-4C41-A161-26511CB7848A}

Now, the crash you reported apparently happens because the GitHub API returns `{ "thread": null }` in the GraphQL response, and the success callback in `octo-buffer.lua:632` tries to index `resp.thread` without checking for `nil`/`vim.NIL` first. I've pushed a fix in b3c1cba that adds a `utils.is_blank(resp.thread)` guard so it fails gracefully with "Failed to create thread" instead of crashing.

However, it seems like this feature unfortunately cannot work via the API today. While, as stated in the [blog post](https://github.blog/changelog/2025-09-25-pull-request-files-changed-public-preview-now-supports-commenting-on-unchanged-lines/) you linked in issue #1460, adding comments outside of diff hunks is possible on the "Files changed" tab in the GH web UI, it also states:

> "API support is currently limited, but comments on unchanged lines will be returned by existing APIs and in webhook events."

And nothing seems to have changed since then...

You can read such comments via the API, but cannot create them, and the mutation silently returns `thread: null` for lines beyond the ±3 context range.

I did consider a couple of workarounds, e.g., snapping the comment to the nearest valid diff line, or adding it as a file comment with a "Re line XX:" prefix, but both would place the comment on an unrelated line on GitHub, which would be confusing for other reviewers seeing it in the web UI. Probably not worth the added complexity and documentation burden.

The `resp.thread` nil guard from b3c1cba might still be worth to consider adding in general, as `resp.thread` could be null in edge cases like a PR being merged/closed mid-review, a review being submitted from another session, or GitHub API hiccups. But, not sure how relevant that actually is.

I'd say this PR can be closed for now, as the original error message ("Cannot place comments outside diff hunks") is still correct given the current API constraint. Happy to revisit once GitHub extends API support to match the web UI behavior! And sorry for the noise.

PR #1465, opened 2026-04-09
fix(folds): guard against out-of-range extmark rows in update_details_arrows

### LINKED ISSUE #1464: Error: Invalid 'line': out of range in update_details_arrows on buffer refresh
### Describe the bug

When viewing a PR/issue buffer and refreshing it with `:e`, an intermittent error occurs:

```
Error in CursorMoved Autocommands for "<buffer=1>":
Lua callback: .../lua/octo/folds.lua:261: Invalid 'line': out of range
stack traceback:
    [C]: in function 'nvim_buf_set_extmark'
    .../lua/octo/folds.lua:261: in function 'update_details_arrows'
    .../lua/octo/folds.lua:194: in function <.../lua/octo/folds.lua:193>
```

### To Reproduce

1. Open a PR that contains `<details>` HTML blocks in its body or comments
2. Run `:e` to refresh the buffer
3. The error appears intermittently when the reloaded buffer has a different line count

### Expected behavior

The buffer should refresh without errors.

### Root cause

`update_details_arrows()` is called by the `CursorMoved` autocmd and `on_key` handler during/after buffer reload. When the buffer content changes (e.g. fewer lines after reload), stale extmarks may reference row positions beyond the new buffer line count. `nvim_buf_set_extmark` then fails with "Invalid 'line': out of range".

### Neovim version

NVIM v0.12

christopherklint97:

Fixes a crash in `update_details_arrows` where `nvim_buf_set_extmark` is called with a row index that exceeds the buffer's line count. This happens when refreshing an Octo buffer (`:e`) causes the buffer to reload with fewer lines while stale extmarks still reference the old (now out-of-range) positions.

Fixes #1464

### Describe how you did it

Added a bounds check in `update_details_arrows()` that skips any extmark whose row is >= the buffer's current line count. This is a minimal, defensive guard — it only affects the transient state during buffer reload and does not change behavior in normal operation.

### Describe how to verify it

1. Open a PR that contains `<details>` HTML blocks in its body or comments
2. Run `:e` to refresh the buffer repeatedly
3. Confirm no "Invalid 'line': out of range" error appears

### Special notes for reviews

The fix is a 4-line change (1 line to get the line count, 3 lines for the guard). No new dependencies or behavioral changes for the normal case.

### Checklist

### DIFF
--- a/lua/octo/folds.lua
+++ b/lua/octo/folds.lua
@@ -227,6 +227,7 @@ end
 ---@param bufnr integer
 function M.update_details_arrows(bufnr)
   local details_ns = vim.api.nvim_create_namespace "octo_details_folds"
+  local line_count = vim.api.nvim_buf_line_count(bufnr)
   local extmarks = vim.api.nvim_buf_get_extmarks(bufnr, details_ns, 0, -1, { details = true })
   local arrow_closed = "▶"
   local arrow_open = "▼"
@@ -235,6 +236,9 @@ function M.update_details_arrows(bufnr)
   for _, extmark in ipairs(extmarks) do
     local id = extmark[1]
     local row = extmark[2]
+    if row >= line_count then
+      goto continue
+    end
     local details = extmark[4] ---@type vim.api.keyset.extmark_details?
     if not details then
       goto continue


williambdean:
Thanks for the PR @christopherklint97 !

PR #1470, opened 2026-04-11
fix(fzf-lua): set kind="repo" on preview buffer before render

### LINKED ISSUE #1469: Not a repo buffer assert fires in fzf-lua repo previewer
### Description

`Not a repo buffer` assert fires on every hover in the fzf-lua repo picker preview (e.g. `:Octo repo list`).

### Reproduction

1. `:Octo repo list` (or any command that opens the fzf-lua repo picker).
2. Move the cursor over any entry so the previewer runs.
3. Error toast fires:

```
Not a repo buffer
stack traceback:
  [C]: in function 'assert'
  .../octo.nvim/lua/octo/model/octo-buffer.lua:1007: in function 'repository'
  .../octo.nvim/lua/octo/model/octo-buffer.lua:112: in function 'render_repo'
  .../octo.nvim/lua/octo/pickers/fzf-lua/previewers.lua:334: in function 'cb'
  ...
```

### Root cause

`lua/octo/pickers/fzf-lua/previewers.lua` in `M.repo` constructs the preview buffer with only `bufnr` set:

```lua
local buffer = OctoBuffer:new {
  bufnr = tmpbuf,
}
buffer:configure()
```

With `opts.node = nil` and `opts.kind = nil`, the constructor's final `else` branch in `OctoBuffer:new` (`lua/octo/model/octo-buffer.lua:74-76`) unconditionally sets `this.kind = "reviewthread"`. Later the graphql callback at `previewers.lua:332-334` assigns `buffer.node = resp.data.repository` without touching `buffer.kind`, so when `render_repo()` → `self:repository()` runs, the assert `assert(self:isRepo(), "Not a repo buffer")` added in c0c98e4 fires because `self.kind == "reviewthread"`.

Before c0c98e4 (`chore: improve type ergonomics for OctoBuffer.node`, 2025-07-15), `self:repository()` was `return self.node` with no type check, so this code path silently worked. The previewer has always built the buffer with the wrong `kind`; the assert just made it visible.

### Environment

- `octo.nvim`: master @ `7f0173a` (also reproduces on current HEAD)
- Neovim: 0.12.x
- Picker: `fzf-lua`
- OS: macOS

### Fix

One-line patch in the graphql callback — assign `buffer.kind = "repo"` next to `buffer.node = resp.data.repository`, so the invariant `isRepo()` checks for is restored before `render_repo()` runs. PR incoming.

anirbanchakraborty-dev:
## Summary

Fixes the `Not a repo buffer` assert that fires on every hover in the fzf-lua repo picker preview (`:Octo repo list`).

`M.repo` in `lua/octo/pickers/fzf-lua/previewers.lua` constructs a scratch `OctoBuffer` with only `bufnr` set. `OctoBuffer:new` sees `opts.node == nil` and its final `else` branch assigns `this.kind = "reviewthread"` — so the preview buffer is created with the wrong `kind`. Later the graphql callback assigns `buffer.node = resp.data.repository` without touching `buffer.kind`, then calls `render_repo()`, which routes through `self:repository()` and trips the `assert(self:isRepo(), "Not a repo buffer")` introduced in c0c98e4 (`chore: improve type ergonomics for OctoBuffer.node`, 2025-07-15).

Before c0c98e4, `self:repository()` was `return self.node` with no type check, so the miskind silently "worked". The previewer has always built the buffer with the wrong `kind`; the assert just made it visible.

The fix assigns `buffer.kind = "repo"` right next to `buffer.node = resp.data.repository` in the callback, restoring the invariant `isRepo()` checks for before `render_repo()` runs.

The alternative — passing `kind = "repo"` to `OctoBuffer:new` up at line 321 — is not sufficient on its own, because the constructor's `else` branch still overrides `opts.kind` whenever `opts.node` is nil. Fixing the constructor is a larger, more invasive change; a one-line patch in the callback is the minimal local fix and matches the shape of the existing code.

Fixes #1469.

## Test plan

- [x] `:Octo repo list` and scroll — preview now renders instead of erroring.
- [x] Repo preview content matches what `render_repo` writes for a real repo buffer.
- [x] No change to non-repo pickers (issues, PRs, discussions, releases).

## Diff

One line in one file:

```diff
       if self.preview_bufnr == tmpbuf and vim.api.nvim_buf_is_valid(tmpbuf) then
         local resp = vim.json.decode(output)
         buffer.node = resp.data.repository
+        buffer.kind = "repo"
         buffer:render_repo()
       end
```


williambdean (approved):
LGTM! Thanks

PR #1481, opened 2026-04-20
fix(commands): fix regression due to discussion opening support

PeterCardenas:

`Octo <number> <repo>` was updated to resolve to going to discussion, but it changed a query that was reused in `goto_issue`.
Only that specific usage was updated to handle it, so here we share in a function to make it more clear that the logic is the same.

couldn't find any

### Describe how to verify it

`<leader>gi` on an issue url or short form link


williambdean (approved):
Thanks!

PR #1484, opened 2026-04-20
ci(deps): bump lua ls version and remove unnecessary dep

PeterCardenas:

Attempting to fix CI here by bumping lua-language-server to latest version and also cleaning up some dependencies made unnecessary by newest neovim version (luvit-meta).


williambdean:
Good to go?

PR #1490, opened 2026-04-24
fix(mutations): add marked/unmarked duplicate event fragments

solyshi:
MarkedAsDuplicateEventFragment and UnmarkedAsDuplicateEventFragment were added to fragments.lua in f7e1654 but not included in mutations.lua, causing 'Fragment was used, but not defined' errors on pr create and other mutation calls.


williambdean (approved):
LGTM! Thanks!

solyshi:
Thanks for the review

Issue #1472, opened 2026-04-14
Hardcoded DiffText color in diff reviews doesn't render well on light themes


mumpo:
### Describe what happened and what your expected to happen

Since the introduction of [this PR](https://github.com/pwntester/octo.nvim/pull/1121), the review diff mode uses hardcoded background colors for the `DiffText` highlight group. While this might render well in a dark theme, this renders poorly in light theme.

It should not use hardcoded values, or at least, provide customization options.

### How to reproduce it (as minimally and precisely as possible)

I'm using the `github_light` theme from https://github.com/projekt0n/github-nvim-theme. Start a review of a PR which has changes in existing lines:

<img width="1913" height="162" alt="Image" src="https://github.com/user-attachments/assets/f92262f3-ba8f-4c40-8dd5-fb53bf219772" />

### Tell us your environment

- nvim version: 0.11.6
- octo.nvim commit hash: 5ae580d
- GitHub CLI version (`gh --version`): 2.88.1
- OS: MacOS
- GitHub product (GitHub, GitHub Enterprise, GHEC / See guide to determine [here](https://docs.github.com/en/get-started/using-github-docs/about-versions-of-github-docs#determining-which-github-product-you-use)):

PR #1500, opened 2026-04-27
fix: replace hardcoded diff highlight colors with configurable groups

mumpo:
## 📋  Summary

This PR attempts to solve this issue: https://github.com/pwntester/octo.nvim/issues/1472

Currently, the highlight groups for `DiffText` in review mode are hardcoded, which makes users unable to customize them. This causes an issue on light theme users like myself, since the hardcoded colors were meant to look good only on dark themes.

We can solve this by introducing two new highlight groups:
- `OctoReviewDiffDeleteText`
- `OctoReviewDiffAddText`

Users can configure these highlight groups to customise the review diff appearance.

_Disclaimer: I've used Claude Code to help me solve the issue._

## 📷 Screenshots

| Case | Screenshot |
| - | - |
| Before | Screenshot 2026-04-27 at 10 08 27|
| After (default colors in light theme) | Screenshot 2026-04-27 at 09 32 31|
| After (custom colors in light theme) | Screenshot 2026-04-27 at 09 41 13|
| After (default colors in dark theme) | Screenshot 2026-04-27 at 09 32 19|

### DIFF
--- a/README.md
+++ b/README.md
@@ -839,6 +839,8 @@ Also, you can use [`cmp-emoji`](https://github.com/hrsh7th/cmp-emoji) or [`blink
 | _OctoStateClosedFloat_            | OctoRedFloat       |
 | _OctoStateMergedFloat_            | OctoPurpleFloat    |
 | _OctoStateDraftFloat_             | OctoGreyFloat      |
+| _OctoReviewDiffDeleteText_        | OctoRed            |
+| _OctoReviewDiffAddText_           | OctoGreen          |
 
 The term `GitHub color` refers to the colors used in the WebUI.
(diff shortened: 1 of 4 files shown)


evanstachowiak:
@mumpo, in the interim, I've found even setting DiffText to NONE doesn't solve the problem and the background is still highlighted. Have you found the same, or any other highlight setting in the meantime to turn this off?

PeterCardenas:
i also think it would be better if we went the way of diffview.nvim for a good default as it can be improved, thanks for the contribution

williambdean:
@PeterCardenas would you think this is fine to merge and then we iterate?

PeterCardenas:
> @PeterCardenas would you think this is fine to merge and then we iterate?

yep i think that's fine!

PR #1512, opened 2026-05-09
feat!: add file panel icon callback

### LINKED ISSUE #1511: Support `mini.icons` for file panel icons
request to support `mini.icons` as an alternative icon provider for the review file panel.

One possible method:

```lua
require("octo").setup({
  file_panel = {
    use_icons = true,
    icon_provider = "mini.icons", -- but default is still "nvim-web-devicons"
  },
})
```

Octo owns `mini.icons` integration: config, docs, health checks, fallback behavior, and provider branch in the renderer. The `mini.icons` branch could use this lookup internally:

```lua
local icon, hl = require("mini.icons").get("file", name)
```

a callback would also work:

```lua
require("octo").setup({
  file_panel = {
    use_icons = true,
    get_icon = function(name, ext)
      return require("mini.icons").get("file", name)
    end,
  },
})
```

Octo only owns the callback contract - call `get_icon(name, ext)` and use the returned icon/highlight. User's config owns the provider-specific lookup.

TyceHerrman:

Adds a `file_panel.get_icon(name, ext)` callback so users can provide file panel icons from custom providers like `mini.icons` without Octo owning provider-specific integration.

Fixes #1511

### Describe how you did it

- Added `file_panel.get_icon` to config defaults, type annotations, and validation.
- Updated the review file panel renderer to call the callback when present, with the existing `nvim-web-devicons` path preserved as the default fallback.
- Updated health/docs so `nvim-web-devicons` is not treated as required when a custom icon callback is configured.
- Avoided adding new Plenary-based tests; verification uses direct headless Neovim checks instead.

### Describe how to verify it

- `stylua --check lua/octo/config.lua lua/octo/reviews/renderer.lua lua/octo/health.lua`
- Headless Neovim checks for callback path, fallback path, and `file_panel.get_icon` validation.
- Full lint/test/typecheck workflows are not currently visible on this PR; verified with stylua and direct headless Neovim checks.

### Special notes for reviews

The callback owns provider-specific lookup. Example:

```lua
file_panel = {
  get_icon = function(name, _ext)
    return require("mini.icons").get("file", name)
  end,
}
```

### Checklist

### DIFF
--- a/lua/octo/config.lua
+++ b/lua/octo/config.lua
@@ -63,6 +63,7 @@ local M = {}
 ---@class OctoConfigFilePanel
 ---@field size number
 ---@field use_icons boolean
+---@field get_icon? fun(name: string, ext: string): string?, string?
 
 ---@class OctoConfigUi
 ---@field use_signcolumn boolean
@@ -311,7 +312,8 @@ function M.get_default_values()
     },
     file_panel = {
       size = 10, -- changed files panel rows
-      use_icons = true, -- use web-devicons in file panel (if false, nvim-web-devicons does not need to be installed)
+      use_icons = true, -- show icons in file panel
+      get_icon = nil, -- callback to provide file panel icons and highlights: function(name, ext) return icon, hl end
     },
     colors = { -- used for highlight groups (see Colors section below)
       white = "#ffffff",
@@ -795,6 +797,7 @@ function M.validate_config()
     if validate_type(config.file_panel, "file_panel", "table") then
       validate_type(config.file_panel.size, "file_panel.size", "number")
       validate_type(config.file_panel.use_icons, "file_panel.use_icons", "boolean")
+      validate_type(config.file_panel.get_icon, "file_panel.get_icon", { "nil", "function" })
     end
     validate_aliases()
     validate_pickers()
--- a/lua/octo/config.lua
+++ b/lua/octo/config.lua
@@ -62,8 +62,7 @@ local M = {}
 
 ---@class OctoConfigFilePanel
 ---@field size number
----@field use_icons boolean
----@field get_icon? fun(name: string, ext: string): string?, string?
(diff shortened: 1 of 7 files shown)


lua/octo/config.lua
```diff
@@ -311,7 +312,8 @@ function M.get_default_values()
     },
     file_panel = {
       size = 10, -- changed files panel rows
-      use_icons = true, -- use web-devicons in file panel (if false, nvim-web-devicons does not need to be installed)
+      use_icons = true, -- show icons in file panel
```
williambdean: use_icons seems redundent now. Can we have default callback and nil will turn off if that is actuallly supported behavior

williambdean (commented):
Thanks for PR

lua/octo/config.lua
TyceHerrman: Maybe the cleaner no-overlap version is actually a single option:

`file_panel.icons = true | false | function`

- `true`: use the default `nvim-web-devicons` provider
- `false`: disable file panel icons
- `function(name, ext)`: use a custom provider

I don't think `nil` is a good disable signal. In Lua config tables, `get_icon = nil` is effectively the same as omitting `get_icon`, so it cannot reliably mean “turn icons off”. It also conflicts with the current fallback meaning, where `nil` means “no custom callback, use nvim-web-devicons”.

imho I think the best options are either keep `use_icons = false` as the explicit off switch, or switch to the single `icons = true | false | function`.

williambdean: I like the icons true, false, function idea

TyceHerrman: cool just pushed the update for review

williambdean:
Can we make sure that if this is a breaking change that we include "!" in the title.

It seems like we will raise on the previous configuration with this change.

PR #1530, opened 2026-07-07
fix: ignore stale reload callbacks

### LINKED ISSUE #1528: Reload callback crashes if Octo buffer is closed before `gh` returns
## Summary

Octo can error from a scheduled `gh` callback after reloading an issue/PR buffer if the callback's original target buffer becomes stale before the async request returns.

This is easiest to hit when an external URL router opens or focuses the same GitHub issue/PR while an Octo reload is still pending, but the underlying issue is the async callback assuming its captured `bufnr` is still a valid target.

## Repro

1. Open an Octo issue or PR buffer.
2. Trigger reload (`<C-r>` / `:Octo ... reload`).
3. Before the GitHub request finishes, route/open the same GitHub issue/PR URL again through an external launcher, or otherwise cause the original target buffer/window to be replaced.

## Observed

Neovim reports a scheduled callback error:

```text
[C]: in function 'nvim_buf_call'
.../octo.nvim/lua/octo/init.lua:111
.../octo.nvim/lua/octo/gh/init.lua
```

## Expected

If the async callback no longer has a valid matching target buffer, Octo should skip that stale callback without a visible error. A debug-level log would be useful, but users should only see a warning/error when the current visible buffer could not be updated or a GitHub/API request failed.

## Suspected Cause

`octo.load_buffer()` captures `bufnr`, starts async `gh`, then later calls `vim.api.nvim_buf_call(bufnr, ...)` without checking whether the buffer is still valid/loaded and still the intended Octo buffer.

## Possible Fix

Guard the callback before `nvim_buf_call`:

```lua
if not vim.api.nvim_buf_is_valid(bufnr) or not vim.api.nvim_buf_is_loaded(bufnr) then
  return
end
```

Cursor restore should also be skipped if the original window/buffer is gone

TyceHerrman:

This PR fixes a stale async reload callback race in `octo.load_buffer()`.

If an Octo buffer is closed, unloaded, renamed, or replaced while an async reload is still in flight, the old callback can later try to render into the stale buffer and crash with an invalid buffer error. The callback now silently returns when the original target is no longer valid or no longer matches the captured Octo URI target.

This also adds a vusted regression test so the behavior is checked inside real headless Neovim without adding Plenary coverage for this case.

Fixes #1528

### Describe how you did it

- Capture the original buffer, window, cursor position, and parsed Octo target before starting `M.load()`.
- Before rendering the async callback result, verify that the buffer is still valid, loaded, parseable, and still points at the same repo/kind/id/hostname.
- Skip cursor restore unless the original window is still valid and still displaying the target buffer.
- Add a vusted spec that monkeypatches `octo.load()` and `octo.create_buffer()` to exercise deleted, renamed, valid, and window-changed reload callbacks with real Neovim buffer APIs.
- Add a vusted CI step to the existing test workflow while keeping the current Plenary test step unchanged.

### Describe how to verify it

Verified locally:

```sh
mise exec -- vusted lua/tests/vusted/load_buffer_stale_spec.lua
```

Result:

```text
4 success
```

Also verified workflow YAML parsing locally.

Note: the existing Plenary suite is still run by CI through the existing workflow setup. My local Plenary run did not complete cleanly because the local runtime could not load `plenary/busted`, so CI remains the source of truth for the existing Plenary job.

### Special notes for reviews

This intentionally leaves existing Plenary tests alone, but does not add a new Plenary test for this regression. The project is tracking migration away from Plenary, and this bug depends on real Neovim buffer/window behavior, so the new coverage uses vusted instead.

Stale callbacks are silent by design; this is an expected async race, not a user-facing error.

### Checklist

- [x] Documentation updates in README.md and doc/octo.txt (not applicable)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,6 +24,15 @@ jobs:
         with:
           neovim: true
           version: ${{ matrix.nvim_version }}
+      - name: Install vusted
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y lua5.1 liblua5.1-0-dev luarocks
+          luarocks --lua-version=5.1 install --local vusted
+      - name: Run vusted tests
+        run: |
+          eval "$(luarocks --lua-version=5.1 path --local)"
+          VUSTED_USE_LOCAL=1 vusted lua/tests/vusted/load_buffer_stale_spec.lua
       - name: Restore plenary.nvim from cache
         uses: actions/cache@v6
         with:
--- a/lua/octo/init.lua
+++ b/lua/octo/init.lua
@@ -98,7 +98,8 @@ end
 function M.load_buffer(opts)
   opts = opts or {}
   local bufnr = opts.bufnr or vim.api.nvim_get_current_buf()
-  local cursor_pos = vim.api.nvim_win_get_cursor(0)
+  local winid = vim.api.nvim_get_current_win()
+  local cursor_pos = vim.api.nvim_win_get_cursor(winid)
   local bufname = vim.fn.bufname(bufnr)
   local buffer_info = uri.parse(bufname)
   if buffer_info == nil then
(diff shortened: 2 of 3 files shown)


williambdean (approved):
Looks good, thanks for the PR!

PR #1541, opened 2026-07-29
feat(pr): add :Octo pr branch to copy head or base branch name

nitaicharan:

Octo can already yank a PR's URL (`:Octo pr url`) and head SHA (`:Octo pr sha`), but not its branch names. The head and base branches are rendered as virtual text in the details block, so they cannot be selected or yanked from the buffer.

Reviewers regularly need those names outside the buffer:

- `git diff <base>...<head>`, `git rebase --onto <base>`
- `gh run list --branch <head>`
- pasting a branch name into a comment or chat

This adds a `branch` action to the PR buffer, mirroring the existing copy commands:

```
:Octo pr branch         copies the head branch (default)
:Octo pr branch head    copies the head branch
:Octo pr branch base    copies the base branch
```

The PR options menu (`<CR>` in a PR buffer) gains matching **Copy Head Branch** and **Copy Base Branch** entries, next to the existing Copy URL and Copy SHA.

NONE

This implements https://github.com/pwntester/octo.nvim/discussions/1540. There is no corresponding Issue — the proposal was raised as a Discussion. Happy to open one if you'd prefer the tracker link.

### Describe how you did it

Three small changes:

1. **`lua/octo/utils.lua`** — new `copy_branch(name, register)`, a third sibling to the existing `copy_url` and `copy_sha`, following their exact shape.

2. **`lua/octo/context.lua`** — `create_buffer_wrapper` now returns `function(...)` and calls `cb(buffer, ...)`.

   This was the only non-obvious part. The wrapper previously returned a zero-argument function, so `context.within_pr` silently discarded everything the `:Octo` dispatcher passed — `branch base` would have been indistinguishable from `branch`. The change is invisible to the ~45 existing call sites: every existing callback declares only `buffer`, and Lua discards extra arguments passed to a function that declares fewer parameters. Keymap callbacks are invoked with no arguments at all.

3. **`lua/octo/commands.lua`** — `M.copy_branch` reads `headRefName` / `baseRefName` off the PR node that `buffer:pullRequest()` already returns, so no additional GraphQL query is needed. Registered as `branch` in the `pr` command table beside `sha`.

Plus the `doc/octo.txt` entry and the two `pr_options` menu entries.

Deliberately left out of scope, to keep the change proportionate: picker mappings (they would need head and base variants across `config.lua`, four picker providers and three README blocks), and tab completion for the `head`/`base` argument (`:Octo pr merge <Tab>` offers no strategy completion either — adding it here would be the first per-subcommand arg table in `completion/init.lua`).

### Describe how to verify it

Automated:

```sh
nvim --headless -c "PlenaryBustedDirectory lua/tests/plenary/ {minimal_init = 'lua/tests/minimal_init.vim'}"
```

Manually, in any PR buffer:

| Action | Expected |
| --- | --- |
| `:Octo pr branch` | copies the head branch |
| `:Octo pr branch head` | copies the head branch |
| `:Octo pr branch base` | copies the base branch |
| `:Octo pr branch nonsense` | `Invalid branch: 'nonsense'. Expected 'head' or 'base'` |
| `:Octo pr branch` outside a PR buffer | `Not a Pull Request buffer` |
| `<CR>` → Copy Head Branch / Copy Base Branch | same as above |

Check the result with `"+p` or `:echo getreg('+')` against the branch names shown in the details block.

### Special notes for reviews

- The `context.lua` change is the only one that touches shared code. I'd appreciate a careful look at it, though I believe it is provably safe for existing callers for the reason given above.
- New tests cover the head/base dispatch, the invalid-argument and blank-ref error paths, the non-PR-buffer path, the `:Octo` dispatcher seam end to end, the argument forwarding contract in `context.lua`, and the two new menu entries.
- On the checklist below: `doc/octo.txt` is updated, but README.md is not — it does not enumerate the `:Octo pr` subcommands, so there was nothing to add there. Let me know if you'd like it mentioned somewhere in the README anyway.

### Checklist


williambdean (approved):
Reviewed the full diff and traced every caller path through the  varargs change. All ~45 existing  callbacks declare only a single  parameter — Lua silently discards extra args, so the change is provably safe.

The dispatch flow works correctly:
-  →  →  →  ✓
-  (no arg) →  →  →  defaults to  ✓
- Menu entries call  — same code path ✓

Tests cover dispatch, error paths, blank refs, non-PR buffers, and the menu entries. Good to merge.

williambdean (approved):
Looks good, thanks for the PR!

PR #1563, opened 2026-08-26
feat(pr): list CI checks in the PR details section

jcarlos7121:
Hi!

I found myself going back and forth between the PR buffer and `Octo pr runs` so I thought I might as well already add it to the PR buffer for display, and it also includes a key shortcut to go to the run buffer if placed the cursor on it.

## Summary

The PR buffer's details section only showed the rollup state (`Checks: ✓SUCCESS`), so finding out *which* check failed meant leaving the buffer for `:Octo pr checks`. It now lists the checks inline, GitHub-style:

image

- **Summary line** with counts per outcome, each in its own colour — the terminal equivalent of GitHub's "1 failing, 2 skipped, 18 successful checks".
- **One row per check**: outcome glyph, workflow-qualified name (`workflow / job`, matching how GitHub labels them), and duration for finished checks.
- **Ordered by what needs attention**: failing → running → skipped → passed.
- Lives in the details block, so it folds with the rest of the details and refreshes on reload/poll like every other detail line.

## Implementation

- `statusCheckRollup.contexts(first: 100)` added to the existing `pull_request` query, covering both `CheckRun` (Actions jobs: `status`, `conclusion`, `startedAt`/`completedAt`, `checkSuite.workflowRun.workflow.name`) and `StatusContext` (external statuses: `context`, `state`). **No additional API request** — it rides along with the data the buffer already fetches.
- `writers.build_checks_details(rollup)` is a pure builder returning detail lines, so it is unit-testable without a buffer.
- Outcome mapping: non-completed runs are *running*; `SUCCESS` is *passed*; `FAILURE`/`TIMED_OUT`/`ACTION_REQUIRED`/`STARTUP_FAILURE` are *failed*; `SKIPPED`/`NEUTRAL`/`CANCELLED`/`STALE` are *skipped*. Durations reuse `utils.seconds_between` + `utils.format_seconds`.
- If `contexts` is missing (older GHES, or no checks at all), the previous single rollup line is rendered unchanged.

## Tests

`lua/tests/plenary/checks_section_spec.lua` — 8 tests: no rollup, contexts-unavailable fallback, summary counts, row ordering, workflow-qualified naming, durations only for finished checks, `StatusContext` rows, all-green case. Full suite green (15 spec files).

Verified against live data: an all-green PR renders 6 rows with workflow names resolved, and a 20-check PR renders the skipped-first ordering correctly.

## Possible follow-up

A 20-check repo produces 20 rows, which pushes the PR body down. If that feels long in practice, the natural knob is a config option to cap the list or to collapse the *passed* rows into the summary line only (failing/running/skipped are the actionable ones). Not added yet — deliberately kept configuration-free until the default is proven annoying.

---

## Follow-up in this branch: jump to the check under the cursor

The rows are real buffer lines, so the cursor can address them:

| Mapping | Action |
|---|---|
| `<localleader>gc` (`goto_check`) | open the check the cursor sits on |

- **Actions jobs** open in octo's own workflow run viewer (`octo.workflow_runs.render`, the same view `:Octo pr checks` reaches with `<CR>`) — the run id is parsed out of `detailsUrl`.
- **Other statuses** (external CI, CLA bots) open their `targetUrl` in the browser.
- A check with no link, or a line with no check, just reports that instead of doing something surprising.

`write_details` records a line → check map on the buffer while it writes the detail lines, so the mapping stays exact regardless of how many other detail lines are present.

Tests: `lua/tests/plenary/check_navigation_spec.lua` (5 tests: workflow run, external URL, no link, no check on line, non-PR buffer) plus one added to `checks_section_spec.lua` for the per-row context map.

---

*Split note: the two buffer-navigation shortcuts that were originally on this branch — `gp` to return to the PR from a workflow run, and `<localleader>q` to close an Octo buffer — now live in #13 so this PR stays focused on the checks list.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,7 @@ require"octo".setup {
       goto_file = { lhs = "gf", desc = "go to file" },
       stack_up = { lhs = "]s", desc = "open the next PR up the stack" },
       stack_down = { lhs = "[s", desc = "open the next PR down the stack" },
+      goto_check = { lhs = "<localleader>gc", desc = "open the CI check under the cursor" },
       add_assignee = { lhs = "<localleader>aa", desc = "add assignee" },
       remove_assignee = { lhs = "<localleader>ad", desc = "remove assignee" },
       create_label = { lhs = "<localleader>lc", desc = "create label" },
--- a/README.md
+++ b/README.md
@@ -293,6 +293,7 @@ require"octo".setup {
     use_signcolumn = false, -- show "modified" marks on the sign column
     use_statuscolumn = true, -- show "modified" marks on the status column
     use_foldtext = true,
+    fold_checks = true, -- fold the CI checks list in the PR details, closed by default
   },
   issues = {
     order_by = { -- criteria to sort results of `Octo issue list`
@@ -453,6 +454,7 @@ require"octo".setup {
       stack_up = { lhs = "]s", desc = "open the next PR up the stack" },
       stack_down = { lhs = "[s", desc = "open the next PR down the stack" },
       goto_check = { lhs = "<localleader>gc", desc = "open the CI check under the cursor" },
+      toggle_checks = { lhs = "<localleader>tc", desc = "fold or unfold the CI checks list" },
       add_assignee = { lhs = "<localleader>aa", desc = "add assignee" },
       remove_assignee = { lhs = "<localleader>ad", desc = "remove assignee" },
       create_label = { lhs = "<localleader>lc", desc = "create label" },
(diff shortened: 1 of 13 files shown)


williambdean:
Thanks for the PR. Overall, very nice feature. 

What do you think about case with large amount of runs? Takes up a bit of real estate. Maybe a fold? Also, I'm experiencing the color from the skips not showing up well on a dark background (gruvbox)

jcarlos7121:
@williambdean you are right.

I added a toggle fold for the CI list that is folded by default and you can display it with local leader + tc or just space.
and now the skips should display much better under dark background

williambdean:
Thanks for the edits. This is working well for me. Let's give it a try!

williambdean (approved):

